PR #116, opened 2019-12-17
Timezone

d4straub:
Many thanks to contributing to nf-core/ampliseq!

Please fill in the appropriate checklist below (delete whatever is not relevant). These are the most common things requested on pull requests (PRs).

## PR checklist
 - [ ] If necessary, also make a PR on the [nf-core/ampliseq branch on the nf-core/test-datasets repo]( https://github.com/nf-core/test-datasets/pull/new/nf-core/ampliseq)
 - [ ] Ensure the test suite passes (`nextflow run . -profile test,docker`).
 - [ ] Make sure your code lints (`nf-core lint .`).
 - [x] `README.md` is updated

**Learn more about contributing:** https://github.com/nf-core/ampliseq/tree/master/.github/CONTRIBUTING.md


docs/usage.md
```diff
@@ -188,6 +189,10 @@ Please note the following requirements:
 2. The metadata file has to follow the [QIIME2 specifications](https://docs.qiime2.org/2019.10/tutorials/metadata/)
 3. In case of multiple sequencing runs, specific naming of samples are required, see [here](#--multipleSequencingRuns)
 
+### `--qiime_timezone`
```
thermokarst: Why not just call this parameter "timezone"?

nextflow.config
```diff
@@ -61,6 +62,11 @@ config_profile_contact = false
 config_profile_url = false
 }
 
+//export Time Zone required for QIIME2 2019.10
+env {
+  TZ = params.qiime_timezone
```
thermokarst: Is there any way to just pass the host env vars through? If so, you could grab TZ, TIMEZONE, and any others that might be related.

apeltzer: We tried doing this with the Singularity scope and passing through the environment variables such as TZ, but that didn't work as expected unfortunately :-(

docs/usage.md
apeltzer: It's easier for us to track this back when naming it like this - not sure what you are doing in QIIME2 in the future, so we know why and where this came from initially 👍 E.g. when you at some point decide to remove this functionality, we might just upgrade QIIME2 and then drop the parameter again.

thermokarst: Sure, I totally get that, but the solution proposed here changes the entire env's timezone --- this parameter name implies that in only impacts QIIME 2.

apeltzer: True point - maybe @d4straub can note that in the docs at least :+1:

d4straub: Thanks for your suggestion @thermokarst, I updated the docs.

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@
 
 ### Pipeline Updates
 
-* Update from QIIME2 v2018.6 to v2019.10, including DADA2 v1.6 to DADA2 v1.10
+* Update from QIIME2 v2018.6 to v2019.10, including DADA2 v1.6 to DADA2 v1.10, specification of '--qiiime_timezone' might be required
```
thermokarst:
```suggestion
* Update from QIIME2 v2018.6 to v2019.10, including DADA2 v1.6 to DADA2 v1.10, specification of '--qiime_timezone' might be required
```

nextflow.config
thermokarst: Not to state the obvious, but did you remember to do use the [singularity env var prefixing syntax](https://singularity.lbl.gov/docs-environment-metadata)? `SINGULARITYENV_TZ`

apeltzer: We tried using the https://www.nextflow.io/docs/latest/config.html#config-singularity `envWhitelist` here, that should expose all set ENV parameters to the Singularity container, which didn't help in this case :-(

PR #119, opened 2020-02-12
Optional submission of input read files through manifest file

DiegoBrambilla:
## PR checklist
 - [x] description of changes (with reason)
This revised code will use sample identifiers from read file names that cannot be parsed, or the user wants different sample names than the one that would result from parsing file names. A manifest can be provided through --manifest, containing the sample identifiers associated to each forward and reverse read file paths. No submission of read files with --reads is required this way. The old behavior remains, i.e. if read file names can be parsed the QIIME way using underscores, they can be treated via --reads.
The procedure of writing the manifest for the QIIME import is slightly changed and now takes the sample name ("pair_id") from the trimming process, and this is so both for the "parse file names for samples" and the manifest case and is done by the trimming process forwarding the sample names to its output channel (Less duplication and fewer if clauses that way).
Multiple sequencing runs are not supported by manifest at this stage, opening an issue about that is the suggested way. 
 - [x] add tests
A test case has been provided. `conf/test_manifest.config` will allow to run the minimal test case:
`$ nextflow run nf-core/ampliseq -profile test_manifest,docker`
 - [x] make a PR on the [nf-core/ampliseq branch on the nf-core/test-datasets repo]( https://github.com/nf-core/test-datasets/pull/new/nf-core/ampliseq)
Two files needed to be added to `nf-core/test-datasets/ampliseq`.
`-profile test_manifest` calls to [Manifest.tsv](https://github.com/nf-core/test-datasets/blob/ampliseq/testdata/Manifest.tsv) and a dedicated [Metadata_multi.tsv](https://github.com/nf-core/test-datasets/blob/ampliseq/testdata/Metadata_manifest.tsv), as downstream analyses require a metadata file that matches the new sample IDs introduced by the manifest file.
 - [x] Make sure your code lints (`nf-core lint .`).
 - [x] `README.md` is updated


docs/usage.md
```diff
@@ -139,6 +139,11 @@ If `-profile` is not specified at all the pipeline will be run locally and expec
 * `test`
   * A profile with a complete configuration for automated testing
   * Includes links to test data so needs no other parameters
+* `uppmax`
+  * A generic configuration profile to be used in the HPC environment [UPPMAX](https://www.uppmax.uu.se/)
+  * Pulls software from dockerhub: [`nfcore/ampliseq`](http://hub.docker.com/r/nfcore/ampliseq/)
+  * More on `-profile uppmax` [here](https://github.com/nf-core/configs/blob/master/docs/uppmax.md)
+  * More on how to run `-profile uppmax` offline and how to set user-specific parameters [here](https://nf-co.re/rnafusion/docs/configuration/uppmax)
```
d4straub: If I am not mistaken this is usually not described here.

main.nf
```diff
@@ -28,8 +28,22 @@ def helpMessage() {
 	                                Note: All samples have to be sequenced in one run, otherwise also specifiy "--multipleSequencingRuns"
 	  --FW_primer [str]             Forward primer sequence
 	  --RV_primer [str]             Reverse primer sequence
-	  --metadata [path/to/file]     Path to metadata sheet, when missing most downstream analysis are skipped (barplots, PCoA plots, ...)
-	  --qiime_timezone [str]		Needs to be specified to resolve a timezone error (default: 'Europe/Berlin')
+          --metadata [path/to/file]     Path to metadata sheet, when missing most downstream analysis are skipped (barplots, PCoA plots, ...).
```
d4straub: I find that description too detailed here, wouldn't that be better located in usage.md?

main.nf
```diff
@@ -28,8 +28,22 @@ def helpMessage() {
...
+					Additional columns defining metadata associated with each sample or feature ID are optional.
+					Metadata files are not required to have additional metadata columns, so a file containing only an ID column is a valid QIIME 2 metadata file.
+					It is not recommended to mix sample and feature IDs in a single metadata file; keep sample and feature metadata stored in separate files.
+                                        Identifiers should be 36 characters long or less, and also contain only ASCII alphanumeric characters
+                                        (i.e. in the range of [a-z], [A-Z], or [0-9]), the period (.) character, or the dash (-) character.
+                                        By default all numeric columns, blanks or NA are removed, and only columns with multiple different values but not all unique are selected.
+                                        The columns which are to be assessed can be specified by --metadata_category, see below.
+	  --manifest [path/to/file]You can submit a manifest file as an alternative way to provide input reads. No submission of read files with --reads is required this way.
```
d4straub: Too detailed imho, better in usage.md.

docs/usage.md
d4straub: If I am not mistaken this is usually not described here.

conf/test_manifest.config
```diff
@@ -0,0 +1,25 @@
...
+  max_memory = 6.GB
+  max_time = 48.h
+  // Input data
+  FW_primer = "GTGYCAGCMGCCGCGGTAA"
+  RV_primer = "GGACTACNVGGGTWTCTAAT"
+  classifier = "https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/GTGYCAGCMGCCGCGGTAA-GGACTACNVGGGTWTCTAAT-gg_13_8-85-qiime2_2019.7-classifier.qza"
+  metadata = "https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/Metadata_manifest.tsv"
+  outdir = "./results"
+  temp_dir = "./results/tmp_dir"
+  manifest = "https://github.com/LNUc-EEMiS/test-datasets/raw/ampliseq/testdata/Manifest.tsv"
```
d4straub: link to nf-core/test-datasets instead

main.nf
```diff
@@ -495,12 +525,16 @@ if (!params.Q2imported){
...
 		ch_fastq_trimmed_manifest
-			.map { forward, reverse -> [ forward.drop(forward.findLastIndexOf{"/"})[0], forward, reverse ] } //extract file name
-			.map { name, forward, reverse -> [ name.toString().take(name.toString().indexOf("_")), forward, reverse ] } //extract sample name
-			.map { name, forward, reverse -> [ name +","+ forward + ",forward\n" + name +","+ reverse +",reverse" ] } //prepare basic synthax
+			.map { name, reads ->
+				def sampleID = name 
+				def fwdReads = reads [0]
+				def revReads = reads [1]
+				[ "${sampleID}" +","+ "${fwdReads}" + ",forward\n" + "${sampleID}" +","+ "${revReads}" +",reverse" ]
+			}
```
d4straub: Have to test myself if all that (still) works as expected. Can do that soon-ish.

d4straub: To test whether `--reads` input works as expected I downloaded the test read files into folder data and did:
`nextflow run DiegoBrambilla/ampliseq -profile test,singularity --reads "data" --readPaths false` with the result
```
  ValueError: Sample IDs found in the table are missing in the metadata: {'2a_S115_L001', '2_S115_L001', '1a_S103_L001', '1_S103_L001'}.
  
  Plugin error from taxa:
  
    Sample IDs found in the table are missing in the metadata: {'2a_S115_L001', '2_S115_L001', '1a_S103_L001', '1_S103_L001'}.
  
  See above for debug info.

```
I am pretty certain that comes from your changes in the above mentioned part.

Running `nextflow run DiegoBrambilla/ampliseq -profile test_multi,singularity --reads "data" --readPaths false -resume` all works fine (cause there are no changes).

DiegoBrambilla: Running
`nextflow run DiegoBrambilla/ampliseq -profile test,singularity --reads "data" --readPaths false`
creates a mismatch between the sample IDs provided by the [test manifest](https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/Metadata_manifest.tsv) and those generated when importing a standard input:
```
Channel
			.fromFilePairs( params.reads + params.extension, size: 2 )
			.ifEmpty { exit 1, "Cannot find any reads matching: ${params.reads}${params.extension}\nNB: Path needs to be enclosed in quotes!" }
			.into { ch_read_pairs; ch_read_pairs_fastqc }
```
The latter creates sample IDs like this: "2a_S115_L001", but downstream analyses expect to receive the same ID as the one reported in the test manifest (something like "2a_S115").
Since this conflict does not verifies when providing raw reads (just tested this weekend), I think updating the [metadata file for the test_manifest](https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/Metadata_manifest.tsv) profile will do the trick.

d4straub: Because I did not specify `--manifest` or `- profile test_manifest`, `test_manifest.config` or `Metadata_manifest.tsv` should have nothing to do with this issue? They simply shouldn't be used as far as I understand. Isn't that so?

DiegoBrambilla: Right, `Metadata_manifest.tsv` or `test_manifest.config` have nothing to do with this issue.
My bad, I was not clear.
However, the said mismatch between sample IDs generated by `main.nf` and sample IDs generated by [Metadata.tsv](https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/Metadata.tsv) stands.

> They simply shouldn't be used as far as I understand. Isn't that so?
`- profile test_manifest` uses `test_manifest.config` and `Metadata_manifest.tsv`, but you already know this. So yeah, they are not used when calling `-profile test` .

Aside from including `--manifest false` in the command line (the pipeline will have to skip downstream analyses, I guess),  a simple way to solve the issue would be to update sample IDs in [Metadata.tsv](https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/Metadata.tsv) and [params.ReadPaths](https://github.com/DiegoBrambilla/ampliseq/blob/master/conf/test.config#L24-L29) so that they both have sample IDs that also include info about the lane, like "2a_S115_L001".

d4straub: > Aside from including --manifest false in the command line (the pipeline will have to skip downstream analyses, I guess), a simple way to solve the issue would be to update sample IDs in Metadata.tsv and params.ReadPaths so that they both have sample IDs that also include info about the lane, like "2a_S115_L001".

While changing the IDs in Metadata.tsv certainly will make theabove mentioned test work, it is in my opinion not an improvement to use "2a_S115_L001" instead of "2a" as sample ID. The latter seems much more intuitive to me. Isn't there a way where you could keep `--reads` working as is and just add the `--manifest` without breaking existing function? Or would that be overly complicated? Probably I can attempt to help?

DiegoBrambilla: @d4straub I am grateful for any help you can offer. The present changes to the code are the simplest result I can propose at the best of my current possibilities. 
Since any further improvement of the code would require expertise in groovy that I do not have, I am stuck.
Before discussing the improvements, consider that composite sample IDs are beneficial in some circumstances. For instance, NGI produces raw reads name with this convention:
[POSITION][FLOWCELL]_[NGI-NAME]_[READ]_[LANE].fastq.gz

-     LANE: Sequencing lane that the file originates from
-     POSITION: Position of the flowcell in the sequencer
-     FLOWCELL: Unique flowcell indentifier
-     NGI-NAME: Internal NGI sample indentifier
-     READ: Forward(1) or reverse(2) read indentifier
NGI users detect unique sample IDs by taking "[POSITION][FLOWCELL]_[NGI-NAME]_[READ]" (often you have read pairs that have the same NGI name and differ only by lane or flowcell position). This is actually what started the need for an input manifest file.
About the improvement to the code, I am replicating the same test case you ran and soon will come back with feedback on which part of the code need to be improved.

d4straub: I would like to emphasize that I totally acknowledge the advantage of having a manifest as import option. I just would prefer if it didn't break any existing routines, if possible at all.

Let me know how testing goes.

DiegoBrambilla: Hi, back with something.
The lines [391-394](https://github.com/DiegoBrambilla/ampliseq/blob/master/main.nf#L391-L394) are the ones that create the sample IDs '2a_S115_L001', '2_S115_L001', '1a_S103_L001', '1_S103_L001' (referring to the aforementioned way you tested ampliseq).
You can verify it by running in your local directory that contains `data` this script:
```
#!/usr/bin/env nextflow

params.reads = "data/*_R{1,2}_001.fastq.gz"

Channel
            .fromFilePairs( params.reads, size: 2 )
            .ifEmpty { exit 1, "Cannot find any reads matching: ${params.reads}${params.extension}\nNB: Path needs to be enclosed in quotes!" }
            .set { ch_read_pairs }
            

process foo {
    echo true
    input:
    set sampleId, file(read) from ch_read_pairs

    script:
    """
    echo tell_me --sample $sampleId --reads $read
    """
}
```
these values are carried until lines [526-536](https://github.com/DiegoBrambilla/ampliseq/blob/master/main.nf#L526-L536) and used without any parsing to build the manifest file for QIIME2 later on in the code. These sample IDs are in contrast with the ones in the test manifest file, as agreed on before.
Ideas on how to solve this:

- Change sample IDs in the [manifest](https://github.com/nf-core/test-datasets/raw/ampliseq/testdata/Metadata.tsv) as follows: '2a_S115_L001', '2_S115_L001', '1a_S103_L001', '1_S103_L001' (deprecated by @d4straub )

- introduce a code snippet [here](https://github.com/DiegoBrambilla/ampliseq/blob/master/main.nf#L391-L394) that parses sample IDs in a smart way, similarly to how it was done [here](https://github.com/nf-core/ampliseq/blob/master/main.nf#L497-L499) in the original ampliseq but without stripping everything comes after the first "_" (0 ideas on how to do it)

- Of course, another possibility would be to put lines [526-536](https://github.com/DiegoBrambilla/ampliseq/blob/master/main.nf#L526-L536) under an if clause ( `if (!params.multipleSequencingRuns && params.manifest) {` ) while keeping lines [497-503](https://github.com/nf-core/ampliseq/blob/master/main.nf#L497-L503) that were present before this PR ( putting them under `} else if (!params.manifest && !params.multipleSequencingRuns ) {`

Tell me what you think or if you have other ideas.

d4straub: Of the three options the last one seems most appealing to me (if clauses) because it is easiest to understand and alter when codes needs to be changed in future.

Could you also add somewhere [here](https://github.com/DiegoBrambilla/ampliseq/blob/02564344b7ef09e67818c8cda0a790bba89737d2/main.nf#L187-L202) a sanity test such as 

```
if (params.multipleSequencingRuns && params.manifest) { 
	exit 1, "The manifest file does not support multiple sequencing runs at this point."
}
```

erikrikarddaniel: I'm pitching, in realizing I'm not as into the code as the two of you are, so bear with my potential ignorance.

As I understand it, the problem arises from that populating the ch_read_pairs in lines 391-4 does not parse out the sample names from the filenames the way it is when fetching read files from urls (readPaths()). This has the consequence that the trimming process will not pass on pair_ids/sample names to the ch_fastq_trimmed_manifest, which was rewritten by @DiegoBrambilla not to parse file names again. True?

In the interest of managing future code, @DiegoBrambilla's second option, parsing the file names in lines 391-4 sounds like the most robust way, as the code that writes the QIIME manifest would not have to be duplicated. (I'm of course partial here, as I was the one suggesting this option in the first place, again in the interest of not duplicating code but keeping the if clauses limited to what happens first in the script.)

I'm perfectly willing to accept your judgement, guys! :-)

docs/usage.md
```diff
@@ -194,6 +196,23 @@ Please note the following requirements:
 2. The metadata file has to follow the [QIIME2 specifications](https://docs.qiime2.org/2019.10/tutorials/metadata/)
 3. In case of multiple sequencing runs, specific naming of samples are required, see [here](#--multipleSequencingRuns)
 
+The first column in the metadata file is the identifier (ID) column and defines the sample or feature IDs associated with your study.
+Additional columns defining metadata associated with each sample or feature ID are optional.
+Metadata files are not required to have additional metadata columns, so a file containing only an ID column is a valid QIIME 2 metadata file.
```
d4straub: This might be true, but without additional columns its useless because there are no groupings for additional analysis. But there might be more output from ampliseq, actually I am not sure since I never tested this.

docs/usage.md
```diff
@@ -194,6 +196,23 @@ Please note the following requirements:
...
+It is not recommended to mix sample and feature IDs in a single metadata file; keep sample and feature metadata stored in separate files.
+Identifiers should be 36 characters long or less, and also contain only ASCII alphanumeric characters
+(i.e. in the range of [a-z], [A-Z], or [0-9]), the period (.) character, or the dash (-) character.
+By default all numeric columns, blanks or NA are removed, and only columns with multiple different values but not all unique are selected.
+The columns which are to be assessed can be specified by `--metadata_category`, see below.
```
d4straub: Well, if `--metadata_category` isn't specified than all columns that fit the specification are automatically chosen. Maybe would be better to mention that here as well?

docs/usage.md
```diff
@@ -194,6 +196,23 @@ Please note the following requirements:
...
+
+### `--manifest`
+
+You can submit a manifest file as an alternative way to provide input reads. No submission of read files with --reads is required this way.
+A manifest is a tab-separated file that must have the following labels in this exact order: sampleID, forwardReads, reverseReads.
```
d4straub: tab/comma/space/semicolon/whatever separated?

main.nf
```diff
@@ -30,19 +30,8 @@ def helpMessage() {
...
-                                        The columns which are to be assessed can be specified by --metadata_category, see below.
-	  --manifest [path/to/file]You can submit a manifest file as an alternative way to provide input reads. No submission of read files with --reads is required this way.
-	                                A manifest is a tab-separated file that must have the following labels in this exact order: sampleID, forwardReads, reverseReads.
-	                                The sample identifiers must be listed under sampleID. Paths to forward and reverse reads must be reported under forwardReads and reverseReads,
-	                                respectively. Test this feature by runnig the pipeline with -profile test_manifest. If downstream analyses do not work, skip them (see below).
-	                                Multiple sequencing runs not supported by manifest at this stage. Default is FALSE. 
+	  --manifest [path/to/file]     Path to manifest.tsv table with the following labels in this exact order: sampleID, forwardReads, reverseReads.
```
d4straub: Also here maybe a hint what separator, even if ".tsv" already sounds like a tab.

docs/usage.md
d4straub: There are really a lot more of these profiles but they are not supposed to be mentioned in this document.

DiegoBrambilla:
> Might that work already with --reads as expected?

@d4straub Could you be more specific? You mean on actual data? I quote from above:

> I have also added the sanity check suggested by @d4straub and checked the whole code with different test cases (first I had to clone my ampliseq fork locally, hence the relative paths):
> 
>     sudo nextflow run ../../ampliseq/main.nf -profile docker,test --reads "data" --readPaths false
> 
>     sudo nextflow run ../../ampliseq/main.nf -profile docker,test
> 
>     sudo nextflow run ../../ampliseq/main.nf -profile docker,test_manifest
> 
> The code no longer breaks in any of the cases above.

Plus, I have now run two additional tests:

-`sudo nextflow run ../../ampliseq/main.nf -profile test_multi,docker --manifest Manifest.tsv`
I have downloaded the manifest file used in `test_manifest` to verify the sanity check, worked as expected:
`ERROR ~ The manifest file does not support multiple sequencing runs at this point.`
 
- sudo nextflow run ../../ampliseq/main.nf -profile test_multi,docker
pipeline runs to completion.

d4straub:
Ok, I tested 
```
nextflow run DiegoBrambilla/ampliseq -profile test,singularity --readPaths false
nextflow run DiegoBrambilla/ampliseq -profile test_multi,singularity --reads "data" --readPaths false
nextflow run DiegoBrambilla/ampliseq -profile test_manifest,singularity
```
and it ran through smoothly! Great!

Could you please address the other minor issues above? Best would be just comment on them and/or write when you have changed it.

DiegoBrambilla:
Hi,
I believe I have addressed all the aforementioned requests of improvement of the documentation.
Are we good to go?

PR #137, opened 2020-02-26
Changing resources in base.config

DiegoBrambilla:
## PR checklist
 - [x] description of changes (with reason)
Resource allocation has been changed in `conf/base.config`. Walltime for process `classifier` has been increased from 2 to 12 h. Also, memory for process `make_SILVA_132_16S_classifier` has been increased while reducing the CPUs to 1 core per iteration.
Testing on real-size databases on UPPMAX with:

- `nextflow run nf-core/ampliseq -r dev -profile uppmax --reads 'data' --trunclenf 290 --trunclenr 210 --FW_primer GTGBCAGCMGCCGCGGTAA --RV_primer GGACTACHVGGGTWTCTAAT`
The process `classifier` fails due to insufficient time limit. It is clear the said process is one, if not the, most-time-consuming step of ampliseq. See graph below:
rackham-snic2017-7-419-diegob-12210267

- `nextflow run DiegoBrambilla/ampliseq -r dev -profile uppmax --reads 'data' --trunclenf 290 --trunclenr 210 --FW_primer GTGBCAGCMGCCGCGGTAA --RV_primer GGACTACHVGGGTWTCTAAT`
Process classifier goes to completion after 1 retry. Also, it has been shown that the process `make_SILVA_132_16S_classifier` runs on a single node. in fact, is not parallelizable (see [here](https://forum.qiime2.org/t/feature-classifier-fit-classifier-naive-bayes-optimisation/10680/1)).
rackham-snic2017-7-419-diegob-11906219

 - [ ] If necessary, also make a PR on the [nf-core/ampliseq branch on the nf-core/test-datasets repo]( https://github.com/nf-core/test-datasets/pull/new/nf-core/ampliseq)
 - [ ] Make sure your code lints (`nf-core lint .`).
Changed small markdown typo errrors
 - [ ] `CHANGELOG.md` is updated
 - [ ] `README.md` is updated


alneberg:
Hi @DiegoBrambilla,

Great that you take the time to improve these things! I believe the classifier can be improved further by increasing the memory even more though. We just ran it with the default values and it timed out after 2 hours, but with the addition of:

```
process{ withName: classifier {
    time = 36h
    cpus = 20
    memory = 120.GB
}}
```

It finished in just 16 minutes. I'm afraid I'm not able to get you the same plots you have (even though this was on rackham as well). While this config is probably overkill, there should be some great time savings to be found with an increase in memory.

erikrikarddaniel:
If 120 GiB is required for Uppmax, but too much for others, should we have
this in an Uppmax specific config, or is that not possible? One wouldn't
want to exclude users that do not have 120 GiB RAM.

/D

Den tors 27 feb. 2020 kl 09:56 skrev Johannes Alneberg <
notifications@github.com>:

DiegoBrambilla:
Hi @alneberg ,
Thanks for your interest and the feedback!

Look, producing the said plots is very easy on UPPMAX, look for [jobstats](https://www.uppmax.uu.se/support/user-guides/jobstats-user-guide/).
if the job is running:
```
jobstats -p -r [jobid] [-M cluster_name] > jobstats_running.out
grep “userid” jobstats_running.out
eog **png
```
if the job is completed:
```
jobstats -p -A [project_name] [-M cluster_name] > jobstats_completed.out
grep “userid” jobsatts_completed.out
eog **png
```

alneberg:
@DiegoBrambilla Thanks, and yes, I know. It's just that it wasn't ran with my account. My colleague ran it and she is now on a conference. 
@erikrikarddaniel, Yes that's definitely possible. Even if 120.GB was set in the regular config, other users will be able to limit this with the [`--max_memory` parameter](https://github.com/nf-core/ampliseq/blob/master/docs/usage.md#--max_memory).

I don't think 120.GB is needed however, we just gave the pipeline what we had available, and it worked surprisingly well. I suggest try to run it with your big dataset with a lot of memory and then find the true limit based on the above plots. I can try to help out, but this is my last work day in 5 months or so...

DiegoBrambilla:
I followed @alneberg 's advice on increasing memory and I couldn't produce a plot of the classifier step because it was completed too quickly (3.45 min!)
I will follow suit and include this in the PR:
```
process{ withName: classifier {
    time = 36h
    cpus = 20
    memory = 120.GB
}}
```

Note: there are insitutition specific - pipeline specific configs: https://github.com/nf-core/configs/tree/master/pipeline
One can even check which uppmax cluster the pipeline is running on to adjust e.g. max memory and such.

conf/base.config
```diff
@@ -26,16 +26,16 @@ process {
...
     time = { check_max (12.h * task.attempt, 'time' ) }
   }
 
   //max memory seen yet: 63.GB
   withName: classifier {
-    cpus = { check_max (8 * task.attempt, 'cpus' ) }
-    memory = { check_max (28.GB * task.attempt, 'memory' ) }
-    time = { check_max (2.h * task.attempt, 'time' ) }
+    cpus = { check_max (20 * task.attempt, 'cpus' ) }
+    memory = { check_max (120.GB * task.attempt, 'memory' ) }
```
d4straub: 120GB is too high in my opinion, this is not required and might lead to unnecessary long queuing or even problems when running on hardware that does not support such high values.

d4straub:
If you guys prefer using high RAM instead of being more considerate to other users of the same cluster, I recommend adding `-c moreRAM.config` with moreRAM.config:
```
process {
  withName: classifier {
    cpus = 20
    memory = 120.GB
    time = 100.h
  }
}
```

conf/base.config
ewels: This code is already increasingly with retries is it not?

What's wrong with allowing users to threshold the memory that is requested using the [`--max_memory` parameter](https://github.com/nf-core/ampliseq/blob/master/docs/usage.md#--max_memory)? This seems to be to be the most flexible approach - set the default to be the fastest possible for those who have the resources, but document that it is easy for people to limit this with the above command-line parameter if this proves to be a problem..

d4straub: For me, default settings should be sensible for most of the users. In my experience, 120Gb is not (queing increases, hardware lmitations). Additional settings should be used in the minority of the cases, and that is what 120Gb in this case is, again, in my opinion. Increasing from e.g. 60Gb to 120Gb is a minor increase in speed for very diverse samples (again, minority) but a hassle for many others.

PR #140, opened 2020-03-09
Revert "Changing resources in base.config"

DiegoBrambilla:
Reverts nf-core/ampliseq#137:

After testing the `dev` revision of `nf-core/ampliseq` on UPPMAX following the [PR](https://github.com/nf-core/ampliseq/pull/137), I got the error message from `make_SILVA_132_16S_classifier`: "Batch job submission failed: Requested node configuration is not available".

This is because UPPMAX expects that the amount of memory per each process is provided by the corresponding number of CPUs provided for the same process (while other HPCs relying on SLURM have other behaviours), and I had specified 1 core and 42 GB of RAM for`make_SILVA_132_16S_classifier` in `base.config`.

So the following changes need to be applied:

- request 7 cores corresponding to 42 GB RAM for `make_SILVA_132_16S_classifier`

- set max_cpus to 40 so that classifier can iterate to 256 GB RAM ( max_memory = 256GB)

Users can limit resources anythime through `--max_cpus` and `--max_memory` as discussed before.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
         # Nextflow versions: check pipeline minimum and current latest
         nxf_ver: ['19.10.0', '']
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v1
       - name: Install Nextflow
         run: |
           export NXF_VER=${{ matrix.nxf_ver }}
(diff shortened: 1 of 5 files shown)


d4straub:
This is really inefficient for other systems than yours. I recommend to have a specific customized config for you. `--max_cpus` and `--max_memory` would reduce the cpus and memory of all processes, not only these two. It would be actually also possible to use a cluster specific pipeline specific config so that this is automatized for you.

PR #154, opened 2020-06-17
Env for template 1.9

d4straub:
This is to adjust the environment.yml for the template 1.9 update. Tests will fail.


d4straub:
Updated according to your suggestions.
Conda works, I think I did some time ago also a container, but I'll confirm this and ping you again.

d4straub:
@ggabernet `docker build .` works fine.

Issue #157, opened 2020-07-10
Add process that identifies sequences still containing primers


erikrikarddaniel:
In some cases, sequences contain duplicated primers. Only the first one will be removed by cutadapt and the processed sequence will hence contain a primer sequence. This is potentially an artifact from the PCR step.

Add a process that searches all ASV sequences for primers and output 1) a table (tsv file) that contains affected ASVs and 2) filtered tables containing information about all ASVs except those containing primers.

erikrikarddaniel:
One possibility is to run cutadapt twice. The first time letting sequences through only if they contain the primers, the second time to *not* let sequences through if they still contain primers.

erikrikarddaniel:
There's an `--discard-trimmed` option to `cutadapt` that can be used when cleaning up sequence pairs containing a second primer. @d4straub, this would be easy enough to implement but raises questions:

1. Do we want this step to be non-mandatory? If so, one needs to create another if block...
2. Does the `--retain_untrimmed` option to the workflow make sense?

I'd like to make the second cutadapt run mandatory and skip the `--retain_untrimmed`, to simplify the workflow.

d4straub:
Ad 1: I am unsure how that second `cutadapt` run (`--discard-trimmed`) would effect results. Generally, I agree that reads should be after the first `cutadapt` run without any primer sequences (independent of using `--retain_untrimmed`), so the second `cutadapt` run discarding primer-containing reads should not break the analysis in any case. However, there might be unforseen complications and to keep the pipeline as broad as possible it might be good to make such a second run non-mandatory, but it seems ok to me to make it default.

Ad 2: Yes this option makes sense, because only primer trimmed reads might be available to the pipeline user. I'll like to have that in the pipeline, if at all possible.

d4straub:
https://github.com/nf-core/test-datasets/pull/195 by @emnilsson prepared tests for this issue.

d4straub:
This was addressed by the linked PR.

Issue #147, opened 2020-05-05
Split make_SILVA_132_16S_classifier into sequence extraction and training?


erikrikarddaniel:
Sequence extraction takes forever on the UPPMAX cluster (4-5 days), and uses little memory, whereas the training step requires more memory (20-25 GiB) and takes much shorter (< 3h). To make it possible to set better cpu and memory limits for this, I would therefore suggest that we split this process into two: The first would run the currently first four steps (unzipping, qiime imports and read extraction) and the second would just be the training step.

d4straub:
Is this still the case (`1.1.2`/`dev`)? The whole `process make_SILVA_132_16S_classifier` takes <3h for me.

erikrikarddaniel:
Den ons 28 okt. 2020 17:03Daniel Strau
<notifications@github.com> skrev:

> Is this still the case (1.1.2/dev)? The whole process
> make_SILVA_132_16S_classifier takes <3h for me.
>
I should think so, but I haven't tried since I've been running a ready made
classifier for a long time now.

In any case, I can't see that it would hurt to make two processes. Emelie
could do this.

/D

d4straub:
Sure, I have no reservations against splitting the process.

d4straub:
This was solved in the linked PR.

PR #193, opened 2020-11-24
Adding a second cutadapt step and splitting "make_classifier" in two

emnilsson:
Added the optional step `--double_primer` to run double cutadapt steps to remove reads where double primers have been introduced - addresses #157. Also split the process `make_classifier` into `clfr_extract_seq` and `clfr_train` to enable more efficient memory use in cases where this is desired - addresses #147.

## PR checklist

- [ ] If necessary, also make a PR on the [nf-core/ampliseq branch on the nf-core/test-datasets repo](https://github.com/nf-core/test-datasets/pull/new/nf-core/ampliseq)


main.nf
```diff
@@ -729,20 +754,14 @@ if( !params.classifier ){
 			.set { ch_ref_database }		
 	}
 
-	process make_classifier {
-		publishDir "${params.outdir}/DB/", mode: params.publish_dir_mode, 
-		saveAs: {filename -> 
-			if (filename.indexOf("${params.FW_primer}-${params.RV_primer}-${params.dereplication}-classifier.qza") == 0) filename
-			else if(params.keepIntermediates) filename 
-			else null}
+	process clfr_extract_seq {
```
d4straub, line 757: just for ease of reading
```suggestion
	process classifier_extract_seq {
```
if you accept this, please replace all occurrences of `clfr` to `classifier`, e.g. in `clfr_train`, in main.nf, configs etc.

main.nf
```diff
@@ -488,12 +489,25 @@ if (!params.Q2imported){
...
 			"""
 			mkdir -p trimmed
-			cutadapt ${primers} ${discard_untrimmed} ${out_files} ${in_files} \
-				 > cutadapt_log_${pair_id}.txt
+			if [[ $params.double_primer=TRUE && $discard_untrimmed="--discard-untrimmed" ]]; then
+	                        mkdir -p firstcutadapt
+				cutadapt ${primers} ${discard_untrimmed} \
+					${int_out_files} \
+					${in_files} >> cutadapt_log_${pair_id}.txt
+                                cutadapt ${primers} --discard-trimmed \
```
d4straub, line 503: Why do you discard here all sequences that contain a second primer? Does that mean all (or many of) these sequences that contain a second primer are artifacts, i.e. not biological sequences, and should not be used further? Did you confirm this? 

A second occurrence of primers in the dataset means that the PCR went horribly wrong and I can imagine that sequences containing two primers are better discarded, I just want to be certain.

main.nf
```diff
@@ -53,6 +53,7 @@ def helpMessage() {
 	  --min_samples [int]           Filtering low prevalent features from the feature table (default: 1)                   
 
 	Cutoffs:
+          --double_primer               Cutdapt will be run twice, first to remove reads without primers (default), then a second time to remove reads that erroneously contain a second set of primers, not to be used with "--retain_untrimmed"
```
d4straub, line 56: > , not to be used with "--retain_untrimmed"

As user, I interpret this in a way that I may not use the parameter '--retain_untrimmed' with `--double_primer`. And this is not true, is it? I read the code below that the second round of cutadapt will ignore '--retain_untrimmed', so it doesn't matter.

main.nf
```diff
@@ -517,9 +531,19 @@ if (!params.Q2imported){
...
-				-o trimmed/$folder${params.split}${reads[0]} -p trimmed/$folder${params.split}${reads[1]} \
-				${reads[0]} ${reads[1]} > cutadapt_log_${pair_id}.txt
+			if [[ $params.double_primer=TRUE && $discard_untrimmed="--discard-untrimmed" ]]; then
+				mkdir -p firstcutadapt
+				cutadapt -g ${params.FW_primer} -G ${params.RV_primer} ${discard_untrimmed} \
+					-o firstcutadapt/$folder${params.split}${reads[0]} -p firstcutadapt/$folder${params.split}${reads[1]} \
+					${reads[0]} ${reads[1]} >> cutadapt_log_${pair_id}.txt
+				cutadapt -g ${params.FW_primer} -G ${params.RV_primer} --discard-trimmed \
+                                        -o trimmed/$folder${params.split}${reads[0]} -p trimmed/$folder${params.split}${reads[1]} \
+                                        firstcutadapt/$folder${params.split}${reads[0]} firstcutadapt/$folder${params.split}${reads[1]} >> cutadapt_log_${pair_id}.txt
```
d4straub, line 541: [QIIME2](https://forum.qiime2.org/t/identical-asvs-with-and-without-adapter-from-cutadapt-dada2/17512/4) has an option `-p-twice` for cutadapt. I am not suggesting to use cutadapt in QIIME2, but probably QIIME2 is using a native parameter of cutadapt to remove double primers? If yes, it would make the task simpler.

nextflow_schema.json
```diff
@@ -76,6 +76,11 @@
                     "description": "Cutadapt will retain untrimmed reads, choose only if input reads are not expected to contain primer sequences.",
                     "help_text": "When read sequences are trimmed, untrimmed read pairs are discarded routinely. Use this option to retain untrimmed read pairs. This is usually not recommended and is only of advantage for specific protocols that prevent sequencing PCR primers. \n\nFor example:\n\n```bash\n--retain_untrimmed\n```"
                 },
+                "double_primer": {
+                    "type": "boolean",
+                    "description": "Cutadapt will be run twice to ensure removal of potential double primers",
+                    "help_text": "Cutdapt will be run twice, first to remove reads without primers (default), then a second time to remove reads that erroneously contain a second set of primers, not to be used with `--retain_untrimmed`"
```
d4straub, line 82: Same as in main.nf: I find the part 
> , not to be used with `--retain_untrimmed`

misleading.

conf/base.config
```diff
@@ -23,12 +23,26 @@ process {
...
   }
 
+  //"make_classifier" is the older version (just one step) of preparing the database for classification
   withName: make_classifier {
     cpus = { check_max (1 * task.attempt, 'cpus' ) }
     memory = { check_max (42.GB * task.attempt, 'memory' ) }
     time = { check_max (12.h * task.attempt, 'time' ) }
   }
 
+  //"clfr_extract_seq" and "clfr_train" is the newer version, where database preparation is split in two
```
d4straub, line 33: As I see it, you deleted `process make_classifier`, so this can be removed here as well.
```suggestion
```

conf/base.config
```diff
@@ -23,12 +23,26 @@ process {
...
+    cpus = { check_max (1 * task.attempt, 'cpus' ) }
+    memory = { check_max (42.GB * task.attempt, 'memory' ) }
+    time = { check_max (12.h * task.attempt, 'time' ) }
+  }
+
+  withName: clfr_train {
+    cpus = { check_max (1 * task.attempt, 'cpus' ) }
+    memory = { check_max (42.GB * task.attempt, 'memory' ) }
+    time = { check_max (12.h * task.attempt, 'time' ) }
+  }
```
d4straub, line 44: Do these have identical resources allocated? Since you argue that you split the process into two, why not immediately adjusting resources?

d4straub (changes requested):
Thanks for contributing!

I have a few comments that would be good to be addressed, I think.

Additionally, while you added a test profile, you did not add it [here](https://github.com/nf-core/ampliseq/blob/master/.github/workflows/ci.yml) to execute it automatically.

main.nf
emnilsson: So far we've only found limited cases with double primers, but in these cases it seems that the PCR did go horribly wrong and keeping these reads is detrimental to the analysis. We haven't been able to think of any cases where a second primer sequence would be of biological origin, but we made this step optional just in case someone can think of a biological reason for double primers.

main.nf
emnilsson: As I've setup the if-else statement - to also evaluate whether "discard untrimmed" is set or not - it won't be compatible. However, it would be possible to change that - but I can't really see that anyone that wants to keep sequences without primers then would like to discard sequences that still contain primers? I thought that "--retain_untrimmed" is for special cases, and not really recommended.

main.nf
emnilsson: I haven't been able to find this option for cutadapt. There is an option to trim away more than one occurrence of the primer, but since we have only seen double primers as a result of bad PCR we want to discard the entire read instead of just trimming away the primer. Therefore I think this is the best approach, but there might be better options.

conf/base.config
emnilsson: I have changed the resource allocation for the ampliseq specific uppmax-config, since that's where I've run this and looked at resource use (and there the first step takes almost 5 days but no memory, and the second step just some hours with more memory). I tried to figure out what to change here, but since the entire process seem to run very fast for you I did not know if the behaviour was the same there. Do you have any idea if the behaviour is similar when you run this? Extracting take a long time but little memory and training is fast with more memory?

main.nf
d4straub: Alright, thanks, have never seen double primers.

main.nf
d4straub: I see. "--retain_untrimmed" exists because the pipeline should not prohibit to analyze reads which had their primers removed previously (but it is indeed not recommended as you mentioned). However, I imagine that also in this cases where previously primers were removed (e.g. a first round of cutadapt), a second primer sequence of a faulty PCR could be retained? But in this case the first round of cutadapt in ampliseq would be the actual second round of primer removal and would need to discard any remaining reads with primers.

Ok, that maybe unnecessarily complicates the matter.

Instead, I am now recommending to add an error message to the pipeline when `--double_primer` and `--retain_untrimmed` are set simultanious. For example in the section [here](https://github.com/nf-core/ampliseq/blob/70b23f4710094df9299e5ab2d18a1d412bbd4d04/main.nf#L181).

main.nf
d4straub: Yes, I see your point regarding removing those reads. I am not aware of a better solution, just came across that info and thought to better share it. Alright, than this is fine with me.

nextflow_schema.json
d4straub: Solved, its not misleading but actually true, as discussed above.

conf/base.config
d4straub: I did not investigate the resources that each step needs in its current form. But before I had this pipeline (and QIIME2 just started), I investigated with bash scripts for each step and sequence extraction needed like no time at all (few minutes) and training ~2.5h). On the two clusters that I used with this pipeline, the combined total time was always ~3h with peak memory usage at ~40GB. So the settings above were totally fine.

So for now these settings are ok, but please remember to review these in case you find a solution for uppmax, that are somewhat reasonable also for other users.

main.nf
emnilsson: To be clear, should I keep it as I set it up and just add the error message?

d4straub: Yes, I think that should do it.
I thought something like
```
if (params.double_primer && params.retain_untrimmed) { 
	exit 1, "Incompatible parameters --double_primer and --retain_untrimmed cannot be both set."
}
```

erikrikarddaniel:
@emnilsson @d4straub I get the following warning when running Emelie's version:

`WARN: There's no process matching config selector: make_classifier -- Did you mean: classifier?`

Might be because there are old config entries mentioning `make_classifier`.

Nice to get rid of, though of course not a show stopper.

emnilsson:
> @emnilsson @d4straub I get the following warning when running Emelie's version:
> 
> `WARN: There's no process matching config selector: make_classifier -- Did you mean: classifier?`
> 
> Might be because there are old config entries mentioning `make_classifier`.
> 
> Nice to get rid of, though of course not a show stopper.

Yes, this has probably to do with the config file which @d4straub pointed out above. I'll remove that!

d4straub (approved):
Ok, this looks good to me!

PR #199, opened 2020-12-11
Change unite flag

emnilsson:
Instead of having `--unite` to specify a unite-database it should now be passed as a string to `--taxon_reference`. `--taxon_reference` only works for silva and unite at the moment.

## PR checklist

- [ ] Documentation in `docs` is updated
- [ ] If necessary, also make a PR on the [nf-core/ampliseq branch on the nf-core/test-datasets repo](https://github.com/nf-core/test-datasets/pull/new/nf-core/ampliseq)

### DIFF
--- a/main.nf
+++ b/main.nf
@@ -67,7 +67,7 @@ def helpMessage() {
 	  --classifier [path/to/file]   Path to QIIME2 classifier file (typically *-classifier.qza)
 	  --classifier_removeHash       Remove all hash signs from taxonomy strings, resolves a rare ValueError during classification (process classifier)
           --reference_database          Path to file with reference database with taxonomies, currently either a qiime compatible file Silva_132_release.zip, or a UNITE fasta file (default: "https://www.arb-silva.de/fileadmin/silva_databases/qiime/Silva_132_release.zip")
-          --unite                       If reference database is in the format of a UNITE fasta file
+          --taxon_reference             Specify which database to use for taxonomic assignment. Either silva or unite (default: silva)
 
 	Statistics:
 	  --metadata_category [str]     Comma separated list of metadata column headers for statistics (default: false)
@@ -215,6 +215,12 @@ if (single_end && !params.manifest) {
 
 if (params.double_primer && params.retain_untrimmed) { 
 	exit 1, "Incompatible parameters --double_primer and --retain_untrimmed cannot be set at the same time."
+}
+//${params.taxon_reference} != 'silva' || ${params.taxon_reference} != 'unite'
+//"--taxon_reference need to be set to either silva or unite"
+System.err.println("${params.taxon_reference}")
+if (!params.classifier){
+        if (!(params.taxon_reference == 'silva' || params.taxon_reference == 'unite')) exit 1, "${params.taxon_reference} ==~ 'silva' || ${params.taxon_reference} ==~ 'unite'"
 }
 
 // AWSBatch sanity checking
@@ -776,7 +782,7 @@ if( !params.classifier ){
 		"""
 		export HOME="\${PWD}/HOME"
 
-		if [ \"${params.unite}\" = \"true\" ]; then
+		if [ ${params.taxon_reference} = "unite" ]; then
                         create_unite_taxfile.py $database db.fa db.tax
 			fasta=\"db.fa\"
 		        taxonomy=\"db.tax\"
--- a/main.nf
+++ b/main.nf
@@ -67,7 +67,7 @@ def helpMessage() {
 	  --classifier [path/to/file]   Path to QIIME2 classifier file (typically *-classifier.qza)
 	  --classifier_removeHash       Remove all hash signs from taxonomy strings, resolves a rare ValueError during classification (process classifier)
           --reference_database          Path to file with reference database with taxonomies, currently either a qiime compatible file Silva_132_release.zip, or a UNITE fasta file (default: "https://www.arb-silva.de/fileadmin/silva_databases/qiime/Silva_132_release.zip")
-          --taxon_reference             Specify which database to use for taxonomic assignment. Either silva or unite (default: silva)
+          --taxon_reference             Specify which database to use for taxonomic assignment. Either 'silva' or 'unite' (default: 'silva')
 
 	Statistics:
 	  --metadata_category [str]     Comma separated list of metadata column headers for statistics (default: false)
@@ -218,7 +218,7 @@ if (params.double_primer && params.retain_untrimmed) {
 }
 
 if (!params.classifier){
-        if (!(params.taxon_reference == 'silva' || params.taxon_reference == 'unite')) exit 1, "--taxon_reference need to be set to either silva or unite"
+        if (!(params.taxon_reference == 'silva' || params.taxon_reference == 'unite')) exit 1, "--taxon_reference need to be set to either 'silva' or 'unite'"
 }
 
 // AWSBatch sanity checking
@@ -780,7 +780,7 @@ if( !params.classifier ){
 		"""
 		export HOME="\${PWD}/HOME"
 
-		if [ ${params.taxon_reference} = "unite" ]; then
+		if [ ${params.taxon_reference} = \"unite\" ]; then
                         create_unite_taxfile.py $database db.fa db.tax
 			fasta=\"db.fa\"
 		        taxonomy=\"db.tax\"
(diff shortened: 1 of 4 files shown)


nextflow_schema.json
```diff
@@ -118,9 +118,11 @@
                     "description": "Path to taxonomic reference database, currently accepts a qiime compatible file Silva_132_release.zip or a UNITE fasta file",
                     "help_text": "By default, the workflow downloads SILVA (https://www.arb-silva.de/) v132 (https://www.arb-silva.de/documentation/release-132/) and extracts reference sequences and taxonomy clustered at 99% similarity and trains a Naive Bayes classifier to assign taxonomy to features."
                 },
-                "unite": {
-                    "type": "boolean",
-                    "description": "If reference database is in the format of a UNITE fasta file instead of SILVA zip file"
+                "taxon_reference": {
+                    "type": "string",
+                    "default": "silva",
+                    "description": "Specify which database to use for taxonomic assignment. Either silva or unite (default: silva).",
```
d4straub, line 124: Please add `"` or `'` around silva / unite

main.nf
```diff
@@ -776,7 +780,7 @@ if( !params.classifier ){
 		"""
 		export HOME="\${PWD}/HOME"
 
-		if [ \"${params.unite}\" = \"true\" ]; then
+		if [ ${params.taxon_reference} = "unite" ]; then
```
d4straub, line 783: So one of the two (old solution or new one) will not work.
Typically `"` needs to be escaped with `\` in a block of string starting & ending with `"""`, such as this block. But if this code works, I wont complain.

main.nf
```diff
@@ -67,7 +67,7 @@ def helpMessage() {
 	  --classifier [path/to/file]   Path to QIIME2 classifier file (typically *-classifier.qza)
 	  --classifier_removeHash       Remove all hash signs from taxonomy strings, resolves a rare ValueError during classification (process classifier)
           --reference_database          Path to file with reference database with taxonomies, currently either a qiime compatible file Silva_132_release.zip, or a UNITE fasta file (default: "https://www.arb-silva.de/fileadmin/silva_databases/qiime/Silva_132_release.zip")
-          --unite                       If reference database is in the format of a UNITE fasta file
+          --taxon_reference             Specify which database to use for taxonomic assignment. Either silva or unite (default: silva)
```
d4straub, line 70: Please add `"` or `'` around silva / unite

d4straub (commented):
The CHANGELOG needs an update as well, I think.

Also, I have the feeling that it would be good to add version numbers to the "silva" and "unite", so that in future we can support multiple versions and just add these. I propose to use here "silva132" and "unite8" or such.

emnilsson:
> The CHANGELOG needs an update as well, I think.
> 
> Also, I have the feeling that it would be good to add version numbers to the "silva" and "unite", so that in future we can support multiple versions and just add these. I propose to use here "silva132" and "unite8" or such.

For the changelog, should I add it to "Added" or "Fixed"?

Sure, sounds good. Do you think we should somehow evaluate them now, or is it fine to just add it to the names?

d4straub:
> For the changelog, should I add it to "Added" or "Fixed"?

I'd say "Added", because it appeared in no release yet, and remove references to `--unite` in the CHANGELOG please

> Sure, sounds good. Do you think we should somehow evaluate them now, or is it fine to just add it to the names?

Evaluate regarding what? Sorry, I do not understand the question.

emnilsson:
> > For the changelog, should I add it to "Added" or "Fixed"?
> 
> I'd say "Added", because it appeared in no release yet, and remove references to `--unite` in the CHANGELOG please
> 
Great, thanks!
> > Sure, sounds good. Do you think we should somehow evaluate them now, or is it fine to just add it to the names?
> 
> Evaluate regarding what? Sorry, I do not understand the question.

The version numbers, sorry, should have been more clear. At some point I assume we want people to be able to say "silva132" and get that version, so I expect that we'll have to check which version is asked for and let ampliseq act accordingly. If I just add "silva132" to the name of the string it would use that version, but it wouldn't be evaluated/checked. Is that fine for now?

main.nf
emnilsson: Hm, I understand what you mean, and for some reason my code worked, with or without escaping. But the old solution is probably more solid, so I'll go with escaping - but also with quotes for the parameter. If that's ok? (Fixing it now.)

d4straub: Not sure if the parameter requires quotation mark now, I'd think it doesn't. Before it did, because true was also in quotation marks.

jtangrot:
> Also, I have the feeling that it would be good to add version numbers to the "silva" and "unite", so that in future we can support multiple versions and just add these. I propose to use here "silva132" and "unite8" or such.

Yes, it makes sense to have version number in there too, esp. if we plan to automatically download the required zip/fasta/taxonomy files. However, if used with --reference_database as it is implemented now for unite, the version is not important as long as the format of the fasta headers is the same. So for unite, it will work for e.g. version 7 too (and I would assume it's the same for other versions of silva if the zip is provided?).

erikrikarddaniel:
Regarding versions, why not emulate e.g. docker and conda and have a non
mandatory version after a separator, i.e. silva=132? Specifying just silva
would give the most recent.

/D

Den fre 11 dec. 2020 09:18emnilsson <notifications@github.com> skrev:

jtangrot:
> edit2: in future "unite8" would ideally immediately download the correct fasta file, as it is done for Silva right now.

That should be quite easy to fix. Is the plan to entirely remove the possibility to use other databases/reference files than those "hardcoded" in the pipeline?

d4straub:
> Yes, it makes sense to have version number in there too, esp. if we plan to automatically download the required zip/fasta/taxonomy files. However, if used with --reference_database as it is implemented now for unite, the version is not important as long as the format of the fasta headers is the same. So for unite, it will work for e.g. version 7 too (and I would assume it's the same for other versions of silva if the zip is provided?).

In terms of Silva, not its not, they had the great idea to change file formats!

> That should be quite easy to fix. Is the plan to entirely remove the possibility to use other databases/reference files than those "hardcoded" in the pipeline?

No, but make it as easy as possible for standard databases (which unite is, I guess). Optimally while keeping parameters to define files yourself. The idea is to keep the required parameters at a minimum but on the other hand allow advanced users to have more control. Not easy of course ;)

> Regarding versions, why not emulate e.g. docker and conda and have a non
mandatory version after a separator, i.e. silva=132? Specifying just silva
would give the most recent.

This sounds great of course!

emnilsson:
> > Regarding versions, why not emulate e.g. docker and conda and have a non
> > mandatory version after a separator, i.e. silva=132? Specifying just silva
> > would give the most recent.
> 
> This sounds great of course!

And in terms of adjusting things now, would we then stick with just "silva" and "unite", or should we then go with "silva=132" and "unite=8"? When we do add more databases I guess either would work, if we will let just "silva" be the latest (supported/available) version.

d4straub:
> And in terms of adjusting things now, would we then stick with just "silva" and "unite", or should we then go with "silva=132" and "unite=8"? When we do add more databases I guess either would work, if we will let just "silva" be the latest (supported/available) version.

Yes, maybe the best way would be to keep it as simple as possible right now, also regarding additional code. So I'd think keeping "silva" and "unite" is fine than, as originally planned. Functionality for specifying the version could be than added in the bigger taxon reference overhaul.

erikrikarddaniel:
Den fre 11 dec. 2020 11:08Daniel Straub <notifications@github.com> skrev:

> And in terms of adjusting things now, would we then stick with just
> "silva" and "unite", or should we then go with "silva=132" and "unite=8"?
> When we do add more databases I guess either would work, if we will let
> just "silva" be the latest (supported/available) version.
>
> Yes, maybe the best way would be to keep it as simple as possible right
> now, also regarding additional code. So I'd think keeping "silva" and
> "unite" is fine than, as originally planned. Functionality for specifying
> the version could be than added in the bigger taxon reference overhaul.
>
I completely agree, just "silva" and "unite" now.

d4straub (approved):
Looks good to me, have you tested if it works as intended?

emnilsson:
> Looks good to me, have you tested if it works as intended?

Well, since it's the classifier step I haven't been able to test as much as I want (very time-consuming). I am running a test with the normal test-profile (without specifying `--classifier`) to see that it behaves as it should all the way through (so that standard silva works). I've tried with a reduced unite-database and that works as expected (at least, all steps finish as they should, haven't verified the output). So, based on that it seems to work. Also, giving a dummy entry for `--taxon_reference` ends in error, as expected.

Issue #180, opened 2020-10-29
MultiQC should not be skipped with --multipleSequencingRuns


d4straub:
MultiQC errors when samples with identical ID are processed from fastQC, this is possible with `--multipleSequencingRuns`, because data has to be structured in different folders. However, this happens rarely.

Better would be to only skip multiQC with `--multipleSequencingRuns` when there are indeed samples with identical ID. 

Solution:
- remove ignoring fastQC error
- remove skipping fastQC when `--multipleSequencingRuns`
- if `--multipleSequencingRuns`, check if sample IDs are all unique, if yes, execute multiQC, if not, skip multiQC

d4straub:
Done.

PR #209, opened 2021-01-27
Edited double primer trimming, improved MultiQC reporting

d4straub:
This solves  #180. 

(1) To allow creating a MultiQC report that summarizes FastQC and Cutadapt results with --multipleSequencingRuns, I edited the processes fastqc_multi and trimming_multi. This was necessary because overlapping IDs were possible. These are (still) not compatible with single_end/PacBio.
(2) I edited processes trimming and trimming_multi with --double_primer so that the cutadapt log shows clear distinction between first and second trimming and that this is also picked up by MultiQC.
(3) I encountered the problem that cutadapt ran twice even if double_primers=false, so I corrected that as well.
(4) I updated the docs accordingly.

edit: I added here @skrakau as reviewer because #180 was her idea and @emnilsson because I changed so much with the double primer trimming. 
edit2: skrakau is in holiday, so I'll skip her approval.


main.nf
```diff
@@ -493,24 +489,32 @@ if (!params.Q2imported){
 			script:
 			discard_untrimmed = params.retain_untrimmed ? '' : '--discard-untrimmed'
 			primers = single_end ? "--rc -g ${params.FW_primer}...${params.RV_primer}" : "-g ${params.FW_primer} -G ${params.RV_primer}"
-			int_out_files = single_end ? "-o firstcutadapt/${reads}" : "-o firstcutadapt/${reads[0]} -p firstcutadapt/${reads[1]}"
-			int_in_files = single_end ? "firstcutadapt/${reads}" : "firstcutadapt/${reads[0]} firstcutadapt/${reads[1]}"
+			in_2_files = single_end ? "second-trimming_${reads}" : "second-trimming_${reads[0]} second-trimming_${reads[1]}" 
+			out_1_files = single_end ? "-o second-trimming_${reads}" : "-o second-trimming_${reads[0]} -p second-trimming_${reads[1]}"
```
emnilsson, line 492: This is a slightly confusing bit (it was before as well, at least to me), but shouldn't these out-files be "first_trimming"? Because I guess it's the out_files (without trimming in the names) that are the files that are used further along?

emnilsson (commented):
In general, I think this looks good. I'd just like clarification about the names in the double cutadapt step.

main.nf
d4straub: Yes, thats quite confusing, I agree. The idea is that the input files to cutadapt are named in a way that describe its process, because the file names are picked up from the log files by MultiQC (and other parsers I guess). 
Meaning, I named the input file for the first round of cutadapt "first-trimming_" (`in_1_files`) and for the second cutadapt round "second-trimming_" (`in_2_files`, which are the output files of the first round: `out_1_files`).
So these names do not describe where they are coming from but where they are going into. I see that this is not intuitive.
If you'd like to see for yourself, just run `nextflow run d4straub/ampliseq -r dev -profile test_doubleprimers,singularity` and let it process until multiqc and look for the MultiQC report. If you have any idea how it might be solved/named better, I am open to ideas.

emnilsson: I see what you mean and looking at the report it looks good.

emnilsson (approved):
I think this looks good - I especially like the solution with using multiqc to report cutadapt statistics.

PR #233, opened 2021-03-22
forward preferentially DADA2 classification to downstream analysis

d4straub:
DADA2 taxonomic classification is forwarded to QIIME2 downstream analysis preferentially over QIIME2 taxonomic classification. QIIME2 taxonomic classification is still in the pipeline but not used further.


erikrikarddaniel (approved):
Looks fine.

PR #238, opened 2021-03-23
Fix forwarding of classification to downstream analysis and support .fna and .fa with --input

d4straub:
- add support for .fna and .fa in addtion to .fasta with --input
- fix linting for nf-core tools 1.13.2
- fix handing over taxonomy classification to downstream analysis


erikrikarddaniel (approved):
LGTM.

PR #245, opened 2021-03-24
Re-add support for addspecies

emnilsson:
When I added the new way of accessing taxonomy-options, I did not think about the addspecies and realised I more or less just excluded it, especially from the test.config (I've re-added it there). But I'm not sure the way I handle it in ampliseq.nf is good - I did the same for `ch_dada_ref_species` as for `ch_dada_ref_taxonomy` but `ch_dada_ref_species` is never used. Maybe just change the `--dada_ref_species` to just evaluate if it's true or not?


conf/test.config
```diff
@@ -23,6 +23,7 @@ params {
   input = "https://github.com/nf-core/test-datasets/raw/ampliseq/samplesheets/Samplesheet.tsv"
   metadata = "https://github.com/nf-core/test-datasets/raw/ampliseq/samplesheets/Metadata.tsv"
   dada_ref_taxonomy = "rdp=18"
+  dada_ref_species = "rdp=18"
```
d4straub, line 26: Since this is only a band-aid on a not required parameter I wait until you have removed it completely.

emnilsson: Yes, working on it!

d4straub (approved):
Looks good!

Issue #244, opened 2021-03-24
md5 sums as keys for sequences


erikrikarddaniel:
I think the md5 sums generated as keys for sequences by QIIME2 was good, and I think we should have that for the DADA2 prepared data as well. I should also add that in the biodiversity project that funds us, we have come to rely on these.

d4straub:
Yes I also prefer md5sums to the arbitrary `ASV_<number>` naming schema that we use now in the DSL2 version.

d4straub:
This was solved in #250 , great!

PR #250, opened 2021-03-30
Unite updates

erikrikarddaniel:
## PR checklist

- [x ] This comment contains a description of changes (with reason).
- [x ] If you've fixed a bug or added code that should be tested, add tests!
 - [ ] If you've added a new tool - add to the software_versions process and a regex to `scrape_software_versions.py`
- [?] Make sure your code lints (`nf-core lint .`).

* @jtangrot wanted both the Unite fungi and Unite all eukaryotes databases from PlutoF, so I added the latter. The Unite databases are now known as "unite-fungi" and "unite-alleuk", with or without "=8.2". I deliberately chose to delete "unite" to make sure one explicitly calls one or the other.
* She also wanted to switch to the "developer" versions, so I did that too. We should remember to document that...

### DIFF
--- a/bin/taxref_reformat_unite.sh
+++ b/bin/taxref_reformat_unite.sh
@@ -7,7 +7,7 @@
 tar xzf *.gz
 
 # Remove leading "k__" and the like and replace space with underscore to create assignTaxonomy.fna
-cat */*.fasta | sed 's/;s__.*//' | sed 's/[a-z]__//g' | sed 's/ /_/g' > assignTaxonomy.fna
+cat */developer/*.fasta | sed 's/;s__.*//' | sed 's/[a-z]__//g' | sed 's/ /_/g' > assignTaxonomy.fna
 
 # Reformat to addSpecies format
 sed 's/>\([^|]\+\)|\([^|]\+|[^|]\+\)|.*/>\2 \1/' assignTaxonomy.fna | sed 's/_/ /g' > addSpecies.fna
--- a/bin/taxref_reformat_unite.sh
+++ b/bin/taxref_reformat_unite.sh
@@ -7,7 +7,7 @@
 tar xzf *.gz
 
 # Remove leading "k__" and the like and replace space with underscore to create assignTaxonomy.fna
-cat */developer/*.fasta | sed 's/;s__.*//' | sed 's/[a-z]__//g' | sed 's/ /_/g' > assignTaxonomy.fna
+cat */developer/*.fasta | sed 's/;[ks]__.*//' | sed 's/[a-z]__//g' | sed 's/ /_/g' > assignTaxonomy.fna
 
 # Reformat to addSpecies format
 sed 's/>\([^|]\+\)|\([^|]\+|[^|]\+\)|.*/>\2 \1/' assignTaxonomy.fna | sed 's/_/ /g' > addSpecies.fna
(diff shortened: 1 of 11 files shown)


erikrikarddaniel:
> Great!
> 
> > She also wanted to switch to the "developer" versions, so I did that too. We should remember to document that...
> 
> Could you please add a sentence to the `nextflow_schema.json` in the help text?
> 
Right. Just a moment and I'll push a commit.

> Also, the "developer" version sounds like it isn't stable. This would be a deal breaker for me because it abandons reproducibility. It would be fine to add an unstable developer version alongside stable databases, but not instead, in my opinion.

AFAIK it is not unstable since it's part of a released set of sequences. "Developer" means that parts of the 18S and 28S sequences are kept, while they are filtered out in the other set -- correct, @jtangrot?

jtangrot:
> > Also, the "developer" version sounds like it isn't stable. This would be a deal breaker for me because it abandons reproducibility. It would be fine to add an unstable developer version alongside stable databases, but not instead, in my opinion.
> 
> AFAIK it is not unstable since it's part of a released set of sequences. "Developer" means that parts of the 18S and 28S sequences are kept, while they are filtered out in the other set -- correct, @jtangrot?

Yes, the reason for choosing the developer version is that any parts of the 18S and 28S sequences are still present in that file while the sequences in the "ordinary" file are cut to only include the ITS part. The sequence set is the same though (i.e. sequence names and number of sequences).

bin/taxref_reformat_unite.sh
```diff
@@ -7,7 +7,7 @@
 tar xzf *.gz
 
 # Remove leading "k__" and the like and replace space with underscore to create assignTaxonomy.fna
-cat */*.fasta | sed 's/;s__.*//' | sed 's/[a-z]__//g' | sed 's/ /_/g' > assignTaxonomy.fna
+cat */developer/*.fasta | sed 's/;[ks]__.*//' | sed 's/[a-z]__//g' | sed 's/ /_/g' > assignTaxonomy.fna
```
jtangrot, line 10: If I understand this correctly, spaces in the actual sequence (not only header) are replaced by underscore. I suggest to remove the spaces in the sequnces instead of replacing them.

erikrikarddaniel: > If I understand this correctly, spaces in the actual sequence (not only header) are replaced by underscore. I suggest to remove the spaces in the sequnces instead of replacing them.

It was meant to be in the names -- easier to get the spaces back with '_' -- and I should change the script accordingly.

d4straub:
> Yes, the reason for choosing the developer version is that any parts of the 18S and 28S sequences are still present in that file while the sequences in the "ordinary" file are cut to only include the ITS part. The sequence set is the same though (i.e. sequence names and number of sequences).

Alright, sounds good then!

Please do not merge before I have managed #232 PR, waiting "only" for passing test runs.

d4straub:
Alright, I changed this PR to nf-core:dev instead of nf-core:DSL2 and removed branch nf-core:DSL2.

jtangrot (approved):
This will leave spaces in some of the sequences, but as assignTaxonomy just ignores them it's not important

erikrikarddaniel:
This addresses issue #244.

PR #257, opened 2021-04-21
Minor changes to make conda work

jtangrot:
Corrected names of conda packages to make it possible to use conda profile. Also removed reference to environment.yml in nextflow.config, as that file is not present/used any more. BTW, the fact that environment.yml is missing results in a lint warning, don't know if that's relevant. The workflow seems to run as it should with  -profile test,conda

## PR checklist

- [X] This comment contains a description of changes (with reason).
 - [ ] If you've added a new tool - add to the software_versions process and a regex to `scrape_software_versions.py`
- [X] Make sure your code lints (`nf-core lint .`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker`).

### DIFF
--- a/modules/local/filter_stats.nf
+++ b/modules/local/filter_stats.nf
@@ -10,7 +10,7 @@ process FILTER_STATS {
         mode: params.publish_dir_mode,
         saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:'') }
 
-    conda (params.enable_conda ? "pandas=1.1.5--py39ha9443f7_0" : null)
+    conda (params.enable_conda ? "pandas=1.1.5=py39ha9443f7_0" : null)
     if (workflow.containerEngine == 'singularity' && !params.singularity_pull_docker_container) {
         container "https://depot.galaxyproject.org/singularity/pandas:1.1.5"
     } else {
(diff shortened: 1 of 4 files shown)


d4straub:
Awesome!

That shines a light on a forgotten problem: the conda packages/containers are actually not yet optimized. Your correction goes into the right direction but isn't perfect yet. The conda package [must not be pinned to a build because they can vary on different platforms](https://github.com/nf-core/modules#software-requirements).

modules/local/filter_stats.nf
```diff
@@ -10,7 +10,7 @@ process FILTER_STATS {
         mode: params.publish_dir_mode,
         saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:'') }
 
-    conda (params.enable_conda ? "pandas=1.1.5--py39ha9443f7_0" : null)
+    conda (params.enable_conda ? "pandas=1.1.5=py39ha9443f7_0" : null)
```
d4straub, line 13: Would you be able to test if the line:
`conda (params.enable_conda ? "pandas=1.1.5" : null)`
works as well, which would follow the guidelines that the build should not be pinned.

d4straub:
> BTW, the fact that environment.yml is missing results in a lint warning, don't know if that's relevant.

That is fine. The linting is still optimized for DSL1 where an `environment.yml` is expected but linting is adapted to raise only a warning to make it compatible with pipelines following the DSL2 scheme.

modules/local/filter_stats.nf
jtangrot: Ahh, didn't know about that. Your suggestion above seems to work!

d4straub (approved):
Great, looks good to me.

PR #299, opened 2021-07-27
Merging template updates 2.0.1

d4straub:
Update with nf-core template update 2.0.1

## PR checklist

- [x] Make sure your code lints (`nf-core lint .`).
- [x] `README.md` is updated (including new tool citations and authors/contributors).


bin/check_samplesheet.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python
```
ggabernet, line 1: This script got added but you are not using it though right? It is still the pipeline default.

bin/parse_samplesheet.nf
```diff
@@ -1,11 +1,11 @@
 // Function to get list of [ meta, [ fastq_1, fastq_2 ] ]
 def parse_samplesheet(LinkedHashMap row, single_end) {
```
ggabernet, line 2: I'm not sure if `bin` is the best location for this function. I think the idea of the `libs/WorkflowAmpliseq` file is exactly to collect these Groovy / nextflow functions. Or won't that work as it uses parameters? In that case I would almost just add it directly into the ampliseq.nf workflow file? Just so that it is easier to locate maybe

modules/nf-core/modules/multiqc/main.nf
```diff
@@ -2,38 +2,34 @@
...
 
 process MULTIQC {
     label 'process_medium'
     publishDir "${params.outdir}",
         mode: params.publish_dir_mode,
-        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), publish_id:'') }
+        saveAs: { filename -> saveFiles(filename:filename, options:params.options, publish_dir:getSoftwareName(task.process), meta:[:], publish_by_meta:[]) }
 
-    conda (params.enable_conda ? "bioconda::multiqc=1.9" : null)
+    conda (params.enable_conda ? "bioconda::multiqc=1.10.1" : null)
```
ggabernet, line 13: MultiQC 1.11 is already available as well, but maybe the module has not been updated yet :)

nextflow.config
```diff
@@ -92,86 +100,76 @@ includeConfig 'conf/base.config'
...
-        podman.enabled = false
-        shifter.enabled = false
-        charliecloud.enabled = false
-        // Avoid this error:
-        //   WARNING: Your kernel does not support swap limit capabilities or the cgroup is not mounted. Memory limited without swap.
-        // Testing this in nf-core after discussion here https://github.com/nf-core/tools/pull/351
-        // once this is established and works well, nextflow might implement this behavior as new default.
-        docker.runOptions = '-u \$(id -u):\$(id -g)'
+        docker.enabled         = true
+        docker.userEmulation   = true
```
ggabernet, line 125:
```suggestion
        docker.runOptions = '-u \$(id -u):\$(id -g)'
```

ggabernet: I know it is in the template, but Paolo said that was not a good option, it will probably be changed again in the new release

subworkflows/local/parse_input.nf
```diff
@@ -4,23 +4,23 @@
 
 params.options = [:]
 
-include { parse_samplesheet } from '../../modules/local/parse_samplesheet' addParams( options: params.options )
+include { parse_samplesheet } from '../../bin/parse_samplesheet' addParams( options: params.options )
```
ggabernet, line 7: Again I'm not sure if bin is the best location. Why not adding the nf code directly as part of the `PARSE_INPUT` subworkflow file?

ggabernet (approved):
Hi @d4straub, looking good, I just have some minor comments!

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Fixed`
 
+* [#299](https://github.com/nf-core/ampliseq/pull/299) - Template update for nf-core/tools version 2.0.1
+
 ### `Dependencies`
```
ggabernet, line 19:
```suggestion
### `Dependencies`
* [#299](https://github.com/nf-core/ampliseq/pull/299) - Updated MultiQC to v1.10
```

d4straub: Right, haven't even realized that update!

nextflow.config
d4straub: Thanks!

modules/nf-core/modules/multiqc/main.nf
d4straub: MultiQC ahs a relatively minor role in ampliseq.

d4straub:
> Also there is the weird .travis.yml.swp file that should be removed :)

I did, thanks!

bin/check_samplesheet.py
d4straub: Yes, I do not use it. Ampliseq parses not just a sample sheet but also allows pointing to a folder or a fasta file with sequences to classify, all with --input. It might be possible to adjust this to the needs here, but `parse_samplesheet.nf ` below does this job already.

bin/parse_samplesheet.nf
d4straub: I'll look into it.

d4straub: I think the correct location would be either `modules/local/functions.nf` or in `subworkflows/local/parse_input.nf`, which parses inputs, including the samplesheet. I chose the latter approach, so that the code for input parsing is less scattered.

subworkflows/local/parse_input.nf
d4straub: Done, right, thats better!

bin/check_samplesheet.py
d4straub: Ok, thats postponed for now.

PR #369, opened 2022-01-21
fix ITSx args

d4straub:
ITSx was not taking any `args` from the `conf/modules.config`. This should fix it.


jtangrot (approved):
Looks good!

d4straub:
Thanks for finding the bug!

Issue #370, opened 2022-01-24
Make addSpecies optional for cut_its too


jtangrot:
## Description of the bug
Maybe not a bug, but shouldn't addSpecies be optional also if you use the cut_its option? In the current dev release, the two opitions "skip_dada_addspecies" and "cut_its" are incompatible.

## Expected behaviour
Accept/implement --skip_dada_addspecies for cut_its too

d4straub:
Agreed, those parameters should work.
The respective code seems to be in [workflows/ampliseq.nf](https://github.com/nf-core/ampliseq/blob/f1abc2010e849903f429a60e130da92b4ad2f8f8/workflows/ampliseq.nf#L319-L334) and adding the code in https://github.com/nf-core/ampliseq/blob/f1abc2010e849903f429a60e130da92b4ad2f8f8/workflows/ampliseq.nf#L321-L324 also to the closure in https://github.com/nf-core/ampliseq/blob/f1abc2010e849903f429a60e130da92b4ad2f8f8/workflows/ampliseq.nf#L327-L333 and adapting `FORMAT_TAXRESULTS` might do the trick.

jtangrot:
Thanks, I'll have a look at it.

d4straub:
Thanks for adding this, I'll close the issue!

PR #372, opened 2022-01-26
Make addSpecies optional also when using cut_its

jtangrot:
## PR checklist

## Comment
This adresses issue #370
Option --skip_dada_addspecies was not compatible with --cut_its, but now this should be fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
+* [#](https://github.com/nf-core/ampliseq/pull/) - Make `--skip_dada_addspecies` compatible with `--cut_its`
 * [#352](https://github.com/nf-core/ampliseq/pull/352) - `--skip_dada_addspecies` allows to skip species level classification to reduce memory requirements, incompatible with `--sbdiexport` and `--cut_its` that expect species annotation
 * [#354](https://github.com/nf-core/ampliseq/pull/354) - Input files and files after primer trimming with cutadapt are required to be >1KB (i.e. not empty) and either the pipeline will stop if at least one sample file fails or the failing samples will be ignored when using `--ignore_empty_input_files` or `--ignore_failed_trimming`, respectively.
 * [#364](https://github.com/nf-core/ampliseq/pull/364) - Adonis in QIIME2 for testing feature importance in beta diversity distances, `--qiime_adonis_formula` can be set to provide a custom formula.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
-* [#](https://github.com/nf-core/ampliseq/pull/) - Make `--skip_dada_addspecies` compatible with `--cut_its`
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
+* [#](https://github.com/nf-core/ampliseq/pull/) - Make `--skip_dada_addspecies` compatible with `--cut_its`
 * [#352](https://github.com/nf-core/ampliseq/pull/352) - `--skip_dada_addspecies` allows to skip species level classification to reduce memory requirements, incompatible with `--sbdiexport` and `--cut_its` that expect species annotation
```
d4straub, line 11: Changes are usually relative to the last release, so those two points could be merged, i.e.
```suggestion
* [#352](https://github.com/nf-core/ampliseq/pull/352), [#372])https://github.com/nf-core/ampliseq/pull/372) - `--skip_dada_addspecies` allows to skip species level classification to reduce memory requirements, incompatible with `--sbdiexport` and `--cut_its` that expect species annotation
```

workflows/ampliseq.nf
```diff
@@ -97,6 +97,7 @@ include { MERGE_STATS                   } from '../modules/local/merge_stats'
 include { DADA2_TAXONOMY                } from '../modules/local/dada2_taxonomy'
 include { DADA2_ADDSPECIES              } from '../modules/local/dada2_addspecies'
 include { FORMAT_TAXRESULTS             } from '../modules/local/format_taxresults'
+include { FORMAT_TAXRESULTS as FORMAT_TAXRESULTS_ADDSP } from '../modules/local/format_taxresults'
```
d4straub, line 100: `FORMAT_TAXRESULTS_ADDSP` should inherit the setting in 
https://github.com/nf-core/ampliseq/blob/f1abc2010e849903f429a60e130da92b4ad2f8f8/conf/modules.config#L226-L232
from `FORMAT_TAXRESULTS` and the files should end up in the `results/dada2` folder. Just want to make sure that this is indeed the case (new syntax...), did you check?

modules/local/format_taxresults.nf
```diff
@@ -8,16 +8,14 @@ process FORMAT_TAXRESULTS {
 
     input:
     path(taxtable)
-    path(taxtable_species)
     path(fastafile)
+    val(outfile)
```
d4straub, line 12: It would be sort of more obvious and functional when
https://github.com/nf-core/ampliseq/blob/f1abc2010e849903f429a60e130da92b4ad2f8f8/bin/add_full_sequence_to_taxfile.py#L35-L39
would actually use the value coming in here I think. But it works the way you are doing it right now I guess.

d4straub (approved):
Looks good to me!
I have only comments/suggestions/questions that are optional, so I approve it right away.

workflows/ampliseq.nf
jtangrot: Yes, I checked and the files end up in the results/dada2 folder as before.

CHANGELOG.md
jtangrot: Good point, I'll fix that.

workflows/ampliseq.nf
d4straub: Great, thanks, always good to learn whether it works as expected!

modules/local/format_taxresults.nf
jtangrot: It works, but you are right. Thanks for the suggestion!

Issue #357, opened 2021-12-02
sample IDs must not start with a number when using metadata


d4straub:
## Is your feature request related to a problem? Please describe

from nf-core slack #ampliseq:

> sample IDs must not start with a number. If they do an X gets added by R, and later QIIME2_BARPLOT can't find them in the metadata. Maybe this could be added to the sample sheet input requirements in the usage docs.

## Describe the solution you'd like

I could imagine 3 ways to resolve this problem:
- Add to documentation. Additionally, perform a test in [parse_input.nf](https://github.com/nf-core/ampliseq/blob/master/subworkflows/local/parse_input.nf) that either produces a warning about downstream analysis problems or even an error.
- Make sure that no X is added by R
- add automatically an X to sample IDs when they start with a number and --metadata is given.

_Edit: that last solution doesnt work, because the metadata is than still without "X" and will not match! SampleIDs also from the metadata needs to be preprended by an X!_

erikrikarddaniel:
> ## Is your feature request related to a problem? Please describe
> 
> from nf-core slack #ampliseq:
> 
> > sample IDs must not start with a number. If they do an X gets added by R, and later QIIME2_BARPLOT can't find them in the metadata. Maybe this could be added to the sample sheet input requirements in the usage docs.
> 
> ## Describe the solution you'd like
> 
> I could imagine 3 ways to resolve this problem:
> 
>     * Add to documentation. Additionally, perform a test in [parse_input.nf](https://github.com/nf-core/ampliseq/blob/master/subworkflows/local/parse_input.nf) that either produces a warning about downstream analysis problems or even an error.
> 
Easy to overlook.

>     * Make sure that no X is added by R
> 
Will require postprocessing of the output file probably and/or ugly if clauses.

>     * add automatically an X to sample IDs when they start with a number and --metadata is given.

I vote for that. Maybe easiest to do even when `--metadata` is not given as it will be added by R anyway?
> 
> ## Describe alternatives you've considered
> ## Additional context

d4straub:
Done in 2.2.0dev

PR #376, opened 2022-01-26
prepend X to sample IDs that start with a number

d4straub:
Closes https://github.com/nf-core/ampliseq/issues/357

I implemented the first proposed solution, but conditional on `--metadata`: When encountering a sample ID that starts with a number the pipeline stops with a descriptive error, but only when metdata is given.

## PR checklist


erikrikarddaniel (approved):
LGTM.

d4straub:
Thanks, but I think I only solved this for the samplesheet input but not for direct fastq file input. Also there sample IDs should be prepended by an X when having a number at the first place. Will do this asap.

d4straub:
Now all sampleIDs derived from samplesheet or direct fastq input will be prepended with an X when starting with a number.

However, after some further test, it appears that this problem is not solved as simple as anticipated. While the sampleIDs are now consistent in the pipeline, the metadata still doesnt match sampleIDs, because it is not modified. Essentially, there would need to be one more process that opens, modifies the metadata ( and potentially verifies the metadata against the sampleIDs from samplesheet/files).
I am going to attempt that soon-ish.

erikrikarddaniel:
I noticed yesterday that you updated the documentation so it actually forbids sample names starting with numbers. Instead of trying to fix sample names, the pipeline could fail with a request to update sample names, stressing that this needs to be done for filenames, input.tsv and metadata. Seems easier.

d4straub:
Ok, I think I'll just throw an error than. But I prefer to only do this in case there is metadata specified. I do not want to burden the user in case there is no metadata, if possible, because the pipeline will succeed in that case regardless of sample IDs.

erikrikarddaniel (approved):
I agree about letting it pass if there's no metadata, so LGTM.

d4straub:
Ok, lets see how it goes and whether there are complains. After all, it would be easy to change again.

PR #397, opened 2022-03-14
add bytesize talk info

d4straub:
This is to add links to the bytesize slides/video and to correct typo's.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,17 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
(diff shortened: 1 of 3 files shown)


marissaDubbelaar (approved):
LGTM

SusiJo (commented):
@d4straub looks good to me! 👍 
Just one question: the new flags that appear in the Changelog are not stated in the docs. 
Wouldn't it make sense to add them in the `usage.md`?

d4straub:
Thanks you two!

@SusiJo You are right, the new parameter and most other parameter are not mentioned in the usage.md. Those are eplained however in [nextflow_schema.json](https://github.com/nf-core/ampliseq/blob/dev/nextflow_schema.json), i.e. https://nf-co.re/ampliseq/dev/parameters. In usage.md describes only the most important aspects, especially input requirements.

Issue #400, opened 2022-03-16
DADA2_QUALITY fails for large number of samples


d4straub:
### Description of the bug

`DADA2_QUALITY` fails for large number of sequences, typically scaling with number of samples (in the range of thousands of samples it fails):
```

Error in sprintf("Total reads: %d", sum(anndf$rc)) : 
    invalid format '%d'; use format %f, %e, %g or %a for numeric objects
  Calls: plotQualityProfile -> annotate -> sprintf
  Execution halted
```

#### Potential solution (from worst to best):
- ignore error of that process
- add --skip_dada_quality
- limit size of table; e.g. subset to first 1000, and give warning that sub setting took place

### Command used and terminal output

```console
any command, where the number of input raw data sequences or number of samples is sufficiently high
```

PR #403, opened 2022-03-17
limit files used for DADA2_QUALITY

d4straub:
Addresses https://github.com/nf-core/ampliseq/issues/400

DADA2_QUALITY file number is limited by total read number to 2147483647. Whenever a set of files is incoming that has more reads, so many files will be dropped until the read number falls below 2147483647. This is only true for that process, any other process is not limited!
This isnt very elegant, but I cannot alter DADA2's code that will fail whenever the read count is above 2147483647 (doesn't fit into INT any more). I tested with the test data, and even when I remove 2 of 4 files, the outcome doesn't change visually or the subsequent calculations.

## PR checklist


modules/local/dada2_quality.nf
```diff
@@ -13,12 +13,61 @@ process DADA2_QUALITY {
...
+    suppressPackageStartupMessages(library(dada2))
+    suppressPackageStartupMessages(library(ShortRead))
+
+    readfiles <- sort(list.files(".", pattern = ".fastq.gz", full.names = TRUE))
+
+    #make list of number of sequences
+    readfiles_length <- countLines(readfiles) / 4
+
+    #use only the first x files when read number gets above 2147483647, read numbers above that do not fit into an INT and crash the process!
+    if ( sum(readfiles_length) > 2147483647 ) {
```
marissaDubbelaar, line 34: Put the numeric value in a vector (since you use it more often than once)

modules/local/dada2_quality.nf
```diff
@@ -13,12 +13,61 @@ process DADA2_QUALITY {
...
+        for (i in length(readfiles):1) {
```
marissaDubbelaar, line 35: Wouldn't it be faster when you use an apply?

modules/local/dada2_quality.nf
```diff
@@ -13,12 +13,61 @@ process DADA2_QUALITY {
...
+    } else {
+        max_files <- length(readfiles)
+        write.table(max_files, file = paste0(max_files," files were used for ${meta} plotQualityProfile.txt"), row.names = FALSE, col.names = FALSE, quote = FALSE, na = '')
+    }
+
+    plot <- plotQualityProfile(readfiles$args)
+    data <- plot\$data
+
+    df <- data.frame(Cycle=character(), Count=character(), Median=character(), stringsAsFactors=FALSE)
+    cycles <- sort(unique(data\$Cycle))
```
marissaDubbelaar, line 52: Is the "\" supposed to be here?

modules/local/dada2_quality.nf
```diff
@@ -13,12 +13,61 @@ process DADA2_QUALITY {
...
+
+    #aggregate data for each sequencing cycle
+    for (cycle in cycles) {
+        subdata <- data[data[, "Cycle"] == cycle, ]
+        score <- list()
+        #convert to list to calculate median
+        for (j in 1:nrow(subdata)) {score <- unlist(c(score, rep(subdata\$Score[j], subdata\$Count[j])))}
```
marissaDubbelaar, line 59: Maybe use a tidy option to make it faster than a loop in  loop

marissaDubbelaar (commented):
Some minor comments, overall it looks good

modules/local/dada2_quality.nf
d4straub: Well, yes, I use it twice! Will do!

modules/local/dada2_quality.nf
d4straub: can I then use `break` (or similar) as below to find the maximum number of files (from first to last)? How would that work?

modules/local/dada2_quality.nf
d4straub: sry, cannot see a `""`, waht do you mean?

modules/local/dada2_quality.nf
marissaDubbelaar: So if I understand correctly, you want to get the highest possible number which is lower than `2147483647` right?
If that is the case, then you don't even need an apply

marissaDubbelaar: `readfiles[which(sum(readfiles_length[1:i]) <= 2147483647 )]` might already do the trick, replacing L35-L40
(don't pin me on this, could not test it and haven't seen the input data 😅 )

modules/local/dada2_quality.nf
marissaDubbelaar: Yea sorry, I meant the slash
I was wondering if `data\$Cycle` needs to be used for using R here or whether it can be `data$Cycle`

modules/local/dada2_quality.nf
d4straub: Hm, not sure we are speaking about the same thing. I have a list of files, I do a list of sequences counts per file. If the total sequence number across all files is above 2147483647, I want to reduce file by file until I reach a total sequence number across all files that is below 2147483647. I do not know how to to this without a loop or such but I am happy to hear a simpler solution.

modules/local/dada2_quality.nf
d4straub: That slash is required because its in the nextflow script. If that would be outside of the nextflow script block, than it may not have a slash at that position, that is true.

marissaDubbelaar (approved):
LGTM

modules/local/dada2_quality.nf
d4straub: Thats great, thanks!

Issue #392, opened 2022-03-08
Add SH assignment for ITS sequences


jtangrot:
### Description of feature

We have been working on how to assign a UNITE Species Hypothesis (SH) to our fungal ITS sequences and come up with a procedure to, in short, match sequences to the database with vsearch and, in case of a good enough match, use the SH of the matching sequence, and replace the taxonomy with that of the SH. Now, we would like to add this to ampliseq, e.g. by adding an option --addsh, as a complement to assignSpecies.

d4straub:
Apologies, I am not familiar with the term _species hypothesis_. A quick google search was not very enlightening. Is that related to https://github.com/nf-core/ampliseq/issues/387? Is it a different kind of database? In any case, it does seem interesting to have it in ampliseq.
However, I would advice to not replace any previous taxonomic classification results but rather add additional files or columns to existing classifications.

jtangrot:
> Apologies, I am not familiar with the term species hypothesis. A quick google search was not very enlightening. Is that related to https://github.com/nf-core/ampliseq/issues/387? Is it a different kind of database?

Some more info on SHs can be found here: https://unite.ut.ee/faq.php
It's not directly related to #387, however it might be possible to extract information on SHs from that too if using the Unite databases. All sequences in the Unite databases are assigned to a SH, but it takes some effort to find which SH at a given sequence identity level, and to extract the taxonomy assignment for that SH.

>  In any case, it does seem interesting to have it in ampliseq.

Great, then I'll go ahead and prepare a PR

> However, I would advice to not replace any previous taxonomic classification results but rather add additional files or columns to existing classifications.

I was planning to create a new file, and replace the taxonomic classification in that file (for entries where the SH can be assigned).

d4straub:
Sounds good!
I hope to have https://github.com/nf-core/ampliseq/pull/385 soon merged that fixes all current linting problems, that would aid further development & review ;)

jtangrot:
Solved in #434

PR #434, opened 2022-05-16
Add module to assign SH to ITS sequences

jtangrot:
This PR adresses issue #392 and adds the option to assign a species hypothesis to the ASVs. 

### Summary of changes:
New modules:
* vsearch_usearch_global.nf - run vsearch with option "usearch_global"
* assignsh.nf - run add_sh_to_taxonomy.py

New files:
* bin/add_sh_to_taxonomy.py - use vsearch results to assign SHs to ASVs

Modified files (other than README, CHANGELOG etc):
* conf/modules.config - config for vsearch_usearch_global and assignsh added
* conf/ref_databases.config - add paths to precalculated sequence-to-SH matching file (seq2sh) and SH taxonomy file (shs.tax)
* workflows/ampliseq.nf - add code for use of assignsh
* CITATIONS.md - add citations to vsearch
* doc/output.md - add output information

## PR checklist


workflows/ampliseq.nf
```diff
@@ -44,12 +44,24 @@ if (params.classifier) {
 
 if (params.dada_ref_taxonomy && !params.skip_taxonomy) {
     ch_dada_ref_taxonomy = Channel.fromList(params.dada_ref_databases[params.dada_ref_taxonomy]["file"]).map { file(it) }
+    if (params.addsh) {
+        if (params.dada_ref_databases[params.dada_ref_taxonomy]["shfile"]) {
+            ch_shinfo = Channel.fromList(params.dada_ref_databases[params.dada_ref_taxonomy]["shfile"]).map { file(it) }
+        } else {
+            exit 1, "SH information is not available for current selection of reference database"
+        }
```
d4straub, line 52: This could be moved to [lib/WorkflowAmpliseq.groovy](https://github.com/nf-core/ampliseq/blob/master/lib/WorkflowAmpliseq.groovy) where typically parameter validation is taken care of.
That should also stop the workflow early so that the parameters are revisited.
Also, I think it would be nice to mention here the relevant parameters `--dada_ref_taxonomy` & `--addsh` and highlighting that this only works with UNITE dbs.

workflows/ampliseq.nf
```diff
@@ -44,12 +44,24 @@ if (params.classifier) {
...
+    }
 } else { ch_dada_ref_taxonomy = Channel.empty() }
 
 if (params.qiime_ref_taxonomy && !params.skip_taxonomy && !params.classifier) {
     ch_qiime_ref_taxonomy = Channel.fromList(params.qiime_ref_databases[params.qiime_ref_taxonomy]["file"]).map { file(it) }
 } else { ch_qiime_ref_taxonomy = Channel.empty() }
 
+if (params.addsh && params.cut_its == "none") {
+    log.warn "Adding SH assignments is only feasible for ITS sequences. Please use option `--cut_its` to find ITS regions in the ASV sequences, unless the given sequences are already cut to the ITS region.\n"
+}
```
d4straub, line 62: That could be (but imho also fine if not) moved to [lib/WorkflowAmpliseq.groovy](https://github.com/nf-core/ampliseq/blob/master/lib/WorkflowAmpliseq.groovy).

workflows/ampliseq.nf
```diff
@@ -366,7 +380,23 @@ workflow AMPLISEQ {
...
+           } else {
+               ch_dada2_tax = DADA2_TAXONOMY.out.tsv
+               if ( params.addsh ) {
+                   VSEARCH_USEARCH_GLOBAL( ch_fasta, ch_assigntax, 'vsearch_blastout.tsv' )
+                   ch_versions = ch_versions.mix(VSEARCH_USEARCH_GLOBAL.out.versions.ifEmpty(null))
+                   ASSIGNSH( ch_dada2_tax, ch_shinfo.collect(), VSEARCH_USEARCH_GLOBAL.out.tsv, 'ASV_tax_SH.tsv')
+                   ch_versions = ch_versions.mix(ASSIGNSH.out.versions.ifEmpty(null))
+                   ch_dada2_tax = ASSIGNSH.out.tsv
+               }
+           }
```
d4straub, line 399: I am not sure whether there isnt potential of bugs when a channel is input and output at once, i.e. here `ch_dada2_tax`. Would this work as well (if yes, than I'd prefer this):
```suggestion
           } else {
               if ( params.addsh ) {
                   VSEARCH_USEARCH_GLOBAL( ch_fasta, ch_assigntax, 'vsearch_blastout.tsv' )
                   ch_versions = ch_versions.mix(VSEARCH_USEARCH_GLOBAL.out.versions.ifEmpty(null))
                   ASSIGNSH( DADA2_TAXONOMY.out.tsv, ch_shinfo.collect(), VSEARCH_USEARCH_GLOBAL.out.tsv, 'ASV_tax_SH.tsv')
                   ch_versions = ch_versions.mix(ASSIGNSH.out.versions.ifEmpty(null))
                   ch_dada2_tax = ASSIGNSH.out.tsv
               } else {
                   ch_dada2_tax = DADA2_TAXONOMY.out.tsv
               }
           }
```
also true for the other addition of `ASSIGNSH` below.

nextflow_schema.json
```diff
@@ -109,6 +109,10 @@
                     "description": "Naming of sequencing files",
                     "help_text": "Indicates the naming of sequencing files (default: `\"/*_R{1,2}_001.fastq.gz\"`).\n\nPlease note:\n\n1. The prepended slash (`/`) is required\n2. The star (`*`) is the required wildcard for sample names\n3. The curly brackets (`{}`) enclose the orientation for paired end reads, seperated by a comma (`,`).\n4. The pattern must be enclosed in quotes\n\nFor example for one sample (name: `1`) with forward (file: `1_a.fastq.gz`) and reverse (file: `1_b.fastq.gz`) reads in folder `data`:\n\n```bash\n--input \"data\" --extension \"/*_{a,b}.fastq.gz\"\n```"
                 },
+                "addsh": {
+                    "type": "boolean",
+                    "description": "If ASVs should be assigned to SHs. Only relevant for ITS data."
```
d4straub, line 114: Just to add here the `species hypothesis` info, not only the abbreviation SH.
```suggestion
                    "description": "If ASVs should be assigned to species hypotheses (SHs). Only relevant for ITS data."
```

d4straub (commented):
Thanks for the addition!
I do have a few comments below.

Also, [prettier](https://prettier.io/) (thats a new code style checker) fails with a typical useless message (see https://github.com/nf-core/ampliseq/runs/6449770921?check_suite_focus=true). Could you run `prettier --write .` locally in the pipeline folder?

bin/add_sh_to_taxonomy.py
```diff
@@ -0,0 +1,148 @@
...
+        exit("Usage: add_sh_to_taxonomy.py <seq2sh.tsv> <SHs.tax> <tax.tsv> <blastout.tab> <outfile> [--species]")
+outfile = sys.argv[5]
+
+# Set number of ranks - include species if --species flag is used
+if with_species:
+    num_ranks = 8
+    tax_entries = ['Domain','Kingdom','Phylum','Class','Order','Family','Genus','Species','SH','confidence']
+else:
+    num_ranks = 7
+    tax_entries = ['Domain','Kingdom','Phylum','Class','Order','Family','Genus','SH','confidence']
```
erikrikarddaniel, line 39: I think we might be forced to be more flexible with ranks, both which and how many, at some point -- could this be read from the `tax.tsv` file header, to make a transition easier? In essence, what you would need in insert an `SH` column as the next to last and `num_ranks` could be counted from the header.

conf/ref_databases.config
```diff
@@ -113,18 +113,21 @@ params {
...
             citation = "Abarenkov, Kessy; Zirk, Allan; Piirmann, Timo; Pöhönen, Raivo; Ivanov, Filipp; Nilsson, R. Henrik; Kõljalg, Urmas (2021): UNITE general FASTA release for Fungi. UNITE Community. 10.15156/BIO/1280049"
             fmtscript = "taxref_reformat_unite.sh"
             dbversion = 'unite-fungi=8.3'
+            shfile = [ "https://scilifelab.figshare.com/ndownloader/files/34497977", "https://scilifelab.figshare.com/ndownloader/files/34497980"]
         }
         'unite-fungi=8.2' {
             title = "UNITE general FASTA release for Fungi - Version 8.2"
             file = [ "https://files.plutof.ut.ee/public/orig/E7/28/E728E2CAB797C90A01CD271118F574B8B7D0DAEAB7E81193EB89A2AC769A0896.gz" ]
             citation = "Abarenkov, Kessy; Zirk, Allan; Piirmann, Timo; Pöhönen, Raivo; Ivanov, Filipp; Nilsson, R. Henrik; Kõljalg, Urmas (2020): UNITE general FASTA release for Fungi. UNITE Community. 10.15156/BIO/786368"
+            shfile = [ "https://scilifelab.figshare.com/ndownloader/files/34497971", "https://scilifelab.figshare.com/ndownloader/files/34497974"]
```
erikrikarddaniel, line 130: Personally, I would be happier if the order of entries were the same -- not a big deal though.

nextflow_schema.json
erikrikarddaniel: I'd say "UNITE species hypotheses (SHs)"

erikrikarddaniel (commented):
* I didn't see a test, or did I miss that?
* I think the vsearch module would be nice as an nf-core module.
* Would it make sense to place all of Unite at Scilife's repo, since we anyway can't add a new version without adding the SH file? In fact, maybe call them 'sbdi-unite' and keep the old Unite dbs as they are?

nextflow_schema.json
d4straub: Sounds good! Also, googling that words lead to a explanatory hit, thats desirable I think ;)

workflows/ampliseq.nf
jtangrot: Thanks for pointing me to that file! Moving the code...

workflows/ampliseq.nf
jtangrot: Agreed that it fits better in lib/WorkflowAmpliseq.groovy

nextflow_schema.json
jtangrot: You're both right, thanks for pointing that out!

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#429](https://github.com/nf-core/ampliseq/pull/429) - `--cutadapt_min_overlap` sets cutadapt's global minimum overlap (`-O`) and `--cutadapt_max_error_rate` sets cutadapt's global maximum error rate (`-e`) for trimming primer sequences.
 - [#431](https://github.com/nf-core/ampliseq/pull/431) - `--skip_dada_quality` allows to skip quality check with DADA2. This is only allowed when `--trunclenf` and `--trunclenr` are set.
+- [#434](https://github.com/nf-core/ampliseq/pull/434) - `--addsh` adds SH to the taxonomy. Only available for UNITE databases.
```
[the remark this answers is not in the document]
jtangrot: Yes, of course, thanks!

d4straub:
Hm the failing linting is caused by the nf-core template update 2.4 https://github.com/nf-core/ampliseq/pull/435
Just ignore it for now.

bin/add_sh_to_taxonomy.py
jtangrot: That's a great suggestion, thanks! I don't know why I didn't do it like that to start with... That also removes the need to check --species flag.

workflows/ampliseq.nf
jtangrot: Thanks, that works and looks better. Changed.

jtangrot:
> * Would it make sense to place all of Unite at Scilife's repo, since we anyway can't add a new version without adding the SH file? In fact, maybe call them 'sbdi-unite' and keep the old Unite dbs as they are?

Well, we can add a new version without the SH files, that will result in an error message that --addsh is not available for that database (might need to rephrase the error message slightly, though). But we could indeed place the reformatted Unite dbs together with the SH files, and remove the need for the reformatting step. But as we always reformat the sequence names in Unite databases maybe we should always use those files for Unite in that case? What do you think @d4straub ?

jtangrot:
> edit: another thought: could you please edit one of the test profiles and include `addsh = true` so that this part of the code is also covered by the CI tests?

Added to the test_pacbio_its profile

d4straub:
> Well, we can add a new version without the SH files, that will result in an error message that --addsh is not available for that database (might need to rephrase the error message slightly, though).

I like that. The error message should be as transparent as possible. However, I do not know whether it is possible to report which ref db versions are actually compatible with SH without hardcoding it in the error message (which will easily be overlooked to amend when adding a new UNITE db version). So the error has to suffice and in worst case it would be trial and error what dbs are compatible with --addsh.

> But we could indeed place the reformatted Unite dbs together with the SH files, and remove the need for the reformatting step. But as we always reformat the sequence names in Unite databases maybe we should always use those files for Unite in that case? What do you think @d4straub ?

I think it would be good to keep the old database entries as they are (adding a file is fine, changing not, imho), to keep backwards comparability as long as possible, e.g. to allow using an older reference database to be able to get comparable taxonomic results also with newer pipeline versions.
Adding a new sbdi-unite might be fine as well. But I do think that most people would tend to welcome an as unchanged UNITE database (i.e. downloaded from UNITE itself as we are doing currently) as possible.

modules/local/assignsh.nf
```diff
@@ -0,0 +1,31 @@
...
+    val  outtable
+
+    output:
+    path outtable        , emit: tsv
+    path "versions.yml"  , emit: versions
+
+    script:
+    def args = task.ext.args ?: ''
+    """
+    add_sh_to_taxonomy.py ${sh_info.join(' ')} $asvtable $blastfile $outtable $args
```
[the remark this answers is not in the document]
jtangrot: > It strikes me that the script might be useful for others as a standalone tool. Would need to be Bioconda packaged though.

Yes, possibly. I did not plan to do that right now, though...

jtangrot:
> I think it would be good to keep the old database entries as they are (adding a file is fine, changing not, imho), to keep backwards comparability as long as possible, e.g. to allow using an older reference database to be able to get comparable taxonomic results also with newer pipeline versions.
> Adding a new sbdi-unite might be fine as well. But I do think that most people would tend to welcome an as unchanged UNITE database (i.e. downloaded from UNITE itself as we are doing currently) as possible.

Then I think I prefer to keep it as is, i.e. to download from Unite and reformat the database both with and without addsh, to avoid storing more or less duplicate information at different places. If you think that's fine @erikrikarddaniel ?

conf/modules.config
```diff
@@ -318,6 +318,27 @@ process {
         ]
     }
 
+    withName: VSEARCH_USEARCHGLOBAL {
+        ext.args = '--top_hits_only --output_no_hits --maxaccepts 50 --query_cov 0.9'
+        publishDir = [
+            [
+                path: { "${params.outdir}/assignsh" },
+                mode: 'copy',
```
d4straub, line 326:
```suggestion
                mode: params.publish_dir_mode,
```

conf/modules.config
```diff
@@ -318,6 +318,27 @@ process {
...
+                pattern: "*.txt"
+            ]
+        ]
+    }
+
+    withName: ASSIGNSH {
+        publishDir = [
+            [
+                path: { "${params.outdir}/assignsh" },
+                mode: 'copy',
```
d4straub, line 336:
```suggestion
                mode: params.publish_dir_mode,
```

docs/output.md
```diff
@@ -146,6 +147,19 @@ Optionally, the ITS region can be extracted from each ASV sequence using ITSx, a
...
+#### assignSH
+
+Optionally, a UNITE species hypothesis (SH) can be added to the taxonomy. In short, the sequences are matched to the selected UNITE database with VSEARCH, using the sequence identity cutoff 98.5%. If a good enough and unique match is found, the sequence is assigned the SH of the matching sequence, and the taxonomy assignment for the sequence is changed to that of the SH. If no match is found, the taxonomy generated by assignTaxonomy and/or assignSpecies is kept.
+
+<details markdown="1">
+<summary>Output files</summary>
+
+- `assignsh/`
+  - `ASV_tax_species_SH.tsv`: Taxonomic classification with SH taxonomy added in case of a match.
+  - `vsearch_blastout.tsv`: Raw vsearch results.
```
d4straub, line 159: Hm and what/where is the `*.txt` file from `VSEARCH_USEARCHGLOBAL`, i.e. https://github.com/jtangrot/ampliseq/blob/59a432805691fbca33a6b924f939fae03de6955a/conf/modules.config#L321-L330?

lib/WorkflowAmpliseq.groovy
```diff
@@ -68,6 +68,20 @@ class WorkflowAmpliseq {
...
         }
 
+        if (params.addsh && !params.dada_ref_databases[params.dada_ref_taxonomy]["shfile"]) {
+            def validDBs = ""
+            for (db in params.dada_ref_databases.keySet()) {
+                if (params.dada_ref_databases[db]["shfile"]) {
+                    validDBs += " " + db
+                }
+            }
+            log.error "UNITE species hypothesis information is not available for the selected reference database, please use the option `--dada_ref_taxonomy` to select an appropriate database. Currently, the option `--addsh` can only be used together with the following UNITE reference databases:\n" + validDBs + "."
```
d4straub, line 78: That seems very nice!

workflows/ampliseq.nf
```diff
@@ -79,6 +83,11 @@ if ( !params.enable_conda && !params.skip_taxonomy && !params.skip_qiime ) {
     run_qiime2 = true
 } else { run_qiime2 = false }
 
+// Set cutoff to use for SH assignment
+if ( params.addsh ) {
+    vsearch_cutoff = 0.985
```
d4straub, line 88: Is that value in any way variable, i.e could there be interest to adjust that? Hard-coded values like this are typically not good. There are exceptions of course.

d4straub (approved):
Looks good to me.
Have a few comments, but I approve it already.

workflows/ampliseq.nf
jtangrot: The reason I put it hard-coded like that is that it's required as input by vsearch_usearchglobal, but we want to keep it the same in ampliseq as SH assignments are are performed using 1.5-SHs (i.e. the user should not change the value). In theory we might want to use other levels for the SHs in the future, but then also other things need to change. I could not figure out a good way to save the value in module config files.

docs/output.md
jtangrot: Thanks, forgot to update that file...

workflows/ampliseq.nf
d4straub: alright, thats fine than.

PR #452, opened 2022-06-24
Add mimetype to input parameter in schema

drpatelh:
This allows using the pipeline with Datasets on Nextflow Tower. We need to be explicit about the mimetype so the Dataset is rendered when launching the pipeline.


d4straub:
Thanks!

> Either a tab-seperated sample sheet, a fasta file, or a folder containing zipped FastQ files

Therefore `"mimetype": "text/tsv",` seems only valid for the `tab-seperated sample sheet`, but not for `fasta file` or `a folder containing zipped FastQ files`?! So would that make fasta & folder input unusable in nextflow tower (would be ok I guess, sample sheet input should be most common input). I guess command line input isn't affected by this (should test that myself)?

drpatelh:
Yep, so adding this `mimetype` will make the pipeline compatible with the Datasets feature on Tower so users can upload and provide samplesheets as input to the pipeline in the UI - they will automatically be rendered in the `input` dropdown box to select. Paths to fasta files and folders will have to be specified manually as normal. 

Command-line input will be unaffected because we aren't using `mimetype` anywhere else.

d4straub (approved):
Great, LGTM!

PR #466, opened 2022-08-18
update credits

d4straub:
Update to the credits in README

@erikrikarddaniel is that fine with you? What affiliation/link do you prefer?


erikrikarddaniel (approved):
Great!

Issue #559, opened 2023-03-27
Add more details on --exclude_taxa value taxonomy format


Ste40:
I would add more information about the format of taxonomy names required by the `--exclude_taxa` parameter.

d4straub:
Hi there, are you speaking about [--exclude_taxa](https://nf-co.re/ampliseq/2.5.0/parameters#exclude_taxa) ? When you click on "help" there is a bunch of information. What is missing there in your opinion?

d4straub:
Thanks for adding the missing explanation yourself!

PR #566, opened 2023-03-29
added description to --exclude_taxa param

dariader:
[Issue 559: Add more details on --exclude_taxa value taxonomy format](https://github.com/nf-core/ampliseq/issues/559)

Added few examples and explanation how to use `--exclude_taxa` parameter from QIIME2 FILTERTAXA method. 

reference docs: https://docs.qiime2.org/2023.2/tutorials/filtering/#filtering-data


d4straub (approved):
Looks good to me.

PR #580, opened 2023-05-09
Feat/init ci nf test

sateeshperi:
Summary 
- ✨ Add NF-TEST pipeline end-to-end tests for existing CI tests
- NF-TEST pipeline Assertions - ✨ Featuring NF-TEST Snapshots, `new File().exists()` and `!new File().exists()` checks
- 🛠️ `tests/pipeline/UTILS.groovy` for local groovy functions used in tests
- 🤖 GitHub Actions CI - pull_request to `dev` tests with NXF_VER `latest-everything`
- 🤖 GitHub Actions CI - pull_request to `master` tests with NXF_VER `22.10.1` & `latest-everything`


tests/pipeline/misc/custom_reftax.nf.test
```diff
@@ -0,0 +1,53 @@
+nextflow_pipeline {
+
+    name "Test Workflow main.nf"
+    script "main.nf"
+    profile "test_reftaxcustom,docker"
```
sateeshperi, line 5: this pipeline uses multiple profiles and I had to place docker along with them in each test file. let me know if you can think of an avoiding this @Emiller88

.github/workflows/ci.yml
```diff
@@ -1,76 +1,178 @@
...
+      - name: Install nf-test
+        run: |
+          wget -qO- https://code.askimed.com/install/nf-test | bash -s ${{ matrix.NFTEST_VER }}
+          sudo mv nf-test /usr/local/bin/
+
+      # Run nf-test with the specified aligner and output test results in TAP format
+      - name: Run nf-test
         run: |
-          nextflow run ${GITHUB_WORKSPACE} -profile test,docker --outdir ./results
+          nf-test test tests/pipeline/${{ matrix.aligner }}/ --tap=test.tap
```
Emiller88, line 85: This should get switched to use tags probably.

tests/pipeline/misc/custom_reftax.nf.test
Emiller88: I've been down this road before 🤠 

```suggestion
    config "config/test_reftaxcustom.config"
```

tests/pipeline/misc/doubleprimers.nf.test
```diff
@@ -0,0 +1,40 @@
+nextflow_pipeline {
+
+    name "Test Workflow main.nf"
+    script "main.nf"
+    profile "test_doubleprimers,docker"
```
Emiller88, line 5:
```suggestion
    config "config/test_doubleprimers.config"
```

Emiller88 (changes requested):
Swap profiles for config paths. Maybe make an issue about migrating to tags?

tests/pipeline/misc/custom_reftax.nf.test
sateeshperi: cool :sunglasses:

sateeshperi: i think we can do this for all the pipelines

sateeshperi: I spoke too fast....I went down this road too....declaring config here overrides the one set in nf-test.config and errors out as it cannot access nextflow.config now @Emiller88 thoughts?

.github/workflows/ci.yml
sateeshperi: i would suggest coming back to using tags once we have explored a common way of implementing them in all pipelines

.github/workflows/ci.yml
```diff
@@ -1,76 +1,115 @@
...
-    if: "${{ github.event_name != 'push' || (github.event_name == 'push' && github.repository == 'nf-core/ampliseq') }}"
+    needs: define_nxf_versions
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        NXF_VER:
-          - "22.10.1"
-          - "latest-everything"
+        NXF_VER: ${{ fromJson(needs.define_nxf_versions.outputs.matrix) }}
+        test_profiles:
```
sateeshperi, line 45: @Emiller88 thoughts on this implementation ? it parallelizes nicely too!

.github/workflows/ci.yml
```diff
@@ -1,76 +1,116 @@
...
+        component:
+          - "tests/pipeline/misc/custom_reftax.nf.test --profile test_reftaxcustom"
+          - "tests/pipeline/misc/doubleprimers.nf.test --profile test_doubleprimers"
+          - "tests/pipeline/misc/fasta_input.nf.test --profile test_fasta"
+          - "tests/pipeline/misc/multi_seq.nf.test --profile test_multi"
+          - "tests/pipeline/paired_end/paired_end.nf.test --profile test"
+          - "tests/pipeline/paired_end/pe_iontorrent.nf.test --profile test_iontorrent"
+          - "tests/pipeline/paired_end/pe_novaseq.nf.test --profile test_novaseq"
+          - "tests/pipeline/paired_end/pe_pacbio_its.nf.test --profile test_pacbio_its"
+          - "tests/pipeline/single_end/single_end.nf.test --profile test_single"
```
d4straub, line 55: Why did you choose 9 out of 11 tests (excluding test_full)?
Is there any reason for renaming them? (not that I care too much about the names...)
And categorizing into misc/pe/se? (not important to me, just to know)

d4straub (commented):
I cannot really comment on most of it.

Did I understand correctly, that:
- `*.nf.test.snap` hold md5sums that are checked for via `*.nf.test` with `snapshot(<file>).match(<key>)`?
- in `*.nf.test`  `new File(<file>).exists()` "only" checks for existence but not checksums

This seems like a good addition.

.github/workflows/ci.yml
sateeshperi: Yes! you are right here. 
> 
    *.nf.test.snap hold md5sums that are checked for via *.nf.test with snapshot(<file>).match(<key>)?
    in *.nf.test new File(<file>).exists() "only" checks for existence but not checksums

1. The decision to test dev PRs with `latest-everything` was made to free up the github runners and test master PR with `22.10.1` and `latest`
2. I didn't know if `test_full` is part of the CI for a PR ? was it previously?
3. agree on test naming...will update them along with changelog
4. the categorization in misc/pe/se was to parallelize tests but we have this now covered in ci `matrix.components` so will move them all to `tests/pipeline`

d4straub: Thanks!
1. fine
2. sorry, no, `test_full` is not and should not run on PR! But 2 other profiles are missing. I checked again, the profiles that you use are complete when looking at the master branch. But dev has two more: [test_pplace.config](https://github.com/nf-core/ampliseq/blob/dev/conf/test_pplace.config) & [test_sintax.config](https://github.com/nf-core/ampliseq/blob/dev/conf/test_sintax.config)
3. thanks
4. alright

sateeshperi: oh missed those....will add them in

sateeshperi: also initially any file that couldn't generate a consistent md5 was moved to exists checks. However we have a helper function to filter lines from a file that go into the snapshot so we can get consistent md5 and also not miss out on imp files integrity checks. if you notice any such files that should be in the snapshot, let me know.

grst (approved):
LGTM! I like how you combine tags and profiles as this allows to run tests with and without nf-test.

.github/workflows/ci.yml
```diff
@@ -1,76 +1,118 @@
...
+          - "test_single"
+          - "test_fasta"
+          - "test_multi"
+          - "test_reftaxcustom"
+          - "test_doubleprimers"
+          - "test_iontorrent"
+          - "test_novaseq"
+          - "test_pacbio_its"
+          - "test_sintax"
+          - "test_pplace"
```
Emiller88, line 57: Just a thought for a future PR, isn't the `test_` a bit redundant?

Emiller88 (approved):
Looks good to me! I feel like the CI is getting complicated, though.

.github/workflows/ci.yml
```diff
@@ -1,76 +1,118 @@
...
+          path: test-datasets/
+          key: ${{ steps.hash_workspace.outputs.digest }}
+
+      - name: Check out test data
+        if: steps.cache-testdata.outputs.cache-hit != 'true'
+        uses: actions/checkout@v3
+        with:
+          repository: nf-core/test-datasets
+          ref: ampliseq
+          path: test-datasets/
```
grst, line 83: Is it really necessary to cache the test data? Maybe just do a shallow checkout every time.

Emiller88: @ewels

.github/workflows/ci.yml
sateeshperi: they need to be the same as profile names

.github/workflows/ci.yml
sateeshperi: ooh that would be nice!

sateeshperi:
> Looks good to me! I feel like the CI is getting complicated, though.

:sweat_smile: on the contrary I feel its streamlining much more now. ok maybe its complicated to beginners....can't decide :shrug:

.github/workflows/ci.yml
grst: ```git clone --depth --branch ampliseq https://github.com/nf-core/test-datasets.git```
takes <2s on my system. I don't think we need that cache. 

Just add 
```yaml
          fetch-depth: 1
```

.github/workflows/ci.yml
d4straub: > I have added in those remaining two tests.

thanks!

> I have tried to snapshot or check for file consistency in all these test profiles but, it is possible I might have missed some that might be important.

Hm, so how are those `*.nf.test.snap` & `*.nf.test` produced? Isnt there some sort of script that goes over the results folder? If yes, how something "important" could be missed? Also asking because I need to maintain that tests somehow (new/changed/reduced files in results)...

tests/pipeline/test.nf.test.snap
```diff
@@ -0,0 +1,76 @@
...
+    },
+    "cutadapt": {
+        "content": [
+            "cutadapt_summary.tsv:md5,5d02749984a811479e7d534fda75163f"
+        ],
+        "timestamp": "2023-05-28T20:55:32+0000"
+    },
+    "software_versions": {
+        "content": [
+            "{BARRNAP={barrnap=0.9}, CUSTOM_DUMPSOFTWAREVERSIONS={python=3.11.0, yaml=6.0}, CUTADAPT_BASIC={cutadapt=3.4}, DADA2_DENOISING={R=4.1.1, dada2=1.22.0}, DADA2_FILTNTRIM={R=4.1.1, dada2=1.22.0}, DADA2_QUALITY1={R=4.1.1, ShortRead=1.52.0, dada2=1.22.0}, DADA2_TAXONOMY={R=4.1.1, dada2=1.22.0}, FASTQC={fastqc=0.11.9}, FILTER_LEN_ASV={Biostrings=2.58.0, R=4.0.3}, FILTER_STATS={pandas=1.1.5, python=3.9.1}, QIIME2_INSEQ={qiime2=2022.11.1}, RENAME_RAW_DATA_FILES={sed=4.7}, SBDIEXPORT={R=3.6.3}, TRUNCLEN={pandas=1.1.5, python=3.9.1}, Workflow={nf-core/ampliseq=2.6.0dev}}"
```
d4straub, line 25: Wouldnt Workflow={nf-core/ampliseq=2.6.0dev} fail whith a new version, should that be in here?

.github/workflows/ci.yml
sateeshperi: > Isnt there some sort of script that goes over the results folder?

No there isn't one yet so, I added all from results folder manually. 

> how something "important" could be missed?

this is in the context of a imp file that didn't make it into the snapshot due to differing md5. however as I said we now have functions to filter lines and then create a snapshot so it gets consistent md5

> nf-test test tests/pipeline/test.nf.test --update-snapshot should do the trick?

Yes, that's it!

.github/workflows/ci.yml
sateeshperi: nice lgtm! ok with switching to this @Emiller88 ?

tests/pipeline/test.nf.test.snap
sateeshperi: Changes to release check list after CI nf-test migration:
- Along with bumping the version number of the dev branch to a release version, also bump the version number of the nf-core/pipeline in snapshots (`Workflow={nf-core/fetchngs=1.11.0dev}}`). This is to ensure that the nf-core/pipeline version in the snapshots is the same as the version in the release branch. Currently to be done by manually deleting the `dev` from the version number in the snapshots. But, in future we can have the tools bump version command to do this for us.

sateeshperi: > Can "content": only have a part of the content or does it need to be complete?

I didn't understand this. could you rephrase?

.github/workflows/ci.yml
d4straub: Thanks, that clarifies it!

tests/pipeline/test.nf.test.snap
d4straub: > Currently to be done by manually deleting the dev from the version number in the snapshots. But, in future we can have the tools bump version command to do this for us.

Great, thanks!

> Can "content": only have a part of the content or does it need to be complete?

rephrase: 
Could we use something like
```
"content": [
            "{BARRNAP={barrnap=0.9}, CUSTOM_DUMPSOFTWAREVERSIONS={python=3.11.0, yaml=6.0}, CUTADAPT_BASIC={cutadapt=3.4}, DADA2_DENOISING={R=4.1.1, dada2=1.22.0}, DADA2_FILTNTRIM={R=4.1.1, dada2=1.22.0}, DADA2_QUALITY1={R=4.1.1, ShortRead=1.52.0, dada2=1.22.0}, DADA2_TAXONOMY={R=4.1.1, dada2=1.22.0}, FASTQC={fastqc=0.11.9}, FILTER_LEN_ASV={Biostrings=2.58.0, R=4.0.3}, FILTER_STATS={pandas=1.1.5, python=3.9.1}, QIIME2_INSEQ={qiime2=2022.11.1}, RENAME_RAW_DATA_FILES={sed=4.7}, SBDIEXPORT={R=3.6.3}, TRUNCLEN={pandas=1.1.5, python=3.9.1}}"
```
e.g. without the `Workflow={nf-core/ampliseq=2.6.0dev}`? Therefore the workflow version wouldn't be needed here (Anyway not needed imho. I'd actually remove the complete test of software_versions, I dont really see why I would need to test that. All that software is containerized, if I intentionally change a container I also intentionally change the software versions).

sateeshperi: We actually remove the nextflow version from the software versions `assert snapshot(UTILS.removeNextflowVersion("$outputDir")).match("software_versions")` so the snapshots can match with different nextflow versions used (22.10.1/latest). That said, I believe recording the software versions including the workflow version in a snapshot is great way to ensure we have a record of the version used for testing and creating the snapshot.

.github/workflows/ci.yml
```diff
@@ -1,76 +1,106 @@
...
       - name: Check out pipeline code
         uses: actions/checkout@v3
 
+      - name: Check out test data
+        uses: actions/checkout@v3
+        with:
+          repository: nf-core/test-datasets
+          ref: ampliseq
+          path: test-datasets/
+          fetch-depth: 1
```
maxulysse, line 71: and so this is enough, and it caches?

sateeshperi: @grst right?

grst: No it doesn't cache, but it **significantly** reduces the download burden by not fetching git history or other branches. In this particular case, doing that clone locally took about 2s.

sateeshperi: but what are we caching actually....just the samplesheets right @maxulysse ?

maxulysse: no, the whole test data, so maybe let's keep the caching but use the shallow git clone

grst: What's the benefit of caching in this case anyway? In both cases it's transferred from a github server on the same network onto the runner VM.

maxulysse: not using the github API to git clone

grst: But that's a super common operation on CI anyway? It literally performs a full clone of the pipeline repo every time the CI gets executed. Did you run into actual API limits or something?

maxulysse: yeaaah, quite often

sateeshperi: I don't think ampliseq datasets are large so, guess we can go with clone for this one?

d4straub (approved):
LGTM

PR #603, opened 2023-06-27
PREPARE: Release 2.6.1

maxulysse:
tiny patches


adamrtalbot (approved):

.github/workflows/ci.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           if [[ "${{ github.event_name }}" == "pull_request" && "${{ github.base_ref }}" == "dev" && "${{ matrix.NXF_VER }}" != "latest-everything" ]]; then
             echo matrix='["latest-everything"]' | tee -a $GITHUB_OUTPUT
           else
-            echo matrix='["latest-everything", "22.10.1"]' | tee -a $GITHUB_OUTPUT
+            echo matrix='["latest-everything", "23.04.0"]' | tee -a $GITHUB_OUTPUT
```
maxulysse, line 35: not caught by tools

maxulysse: cc @mirpedrol

maxulysse: @sateeshperi I'm assuming this is your magic

README.md
```diff
@@ -6,7 +6,7 @@
 [![Cite with Zenodo](http://img.shields.io/badge/DOI-10.5281/zenodo.1493841-1073c8?labelColor=000000)](https://doi.org/10.5281/zenodo.1493841)
 [![Cite Publication](https://img.shields.io/badge/Cite%20Us!-Cite%20Publication-important?labelColor=000000)](https://doi.org/10.3389/fmicb.2020.550420)
 
-[![Nextflow](https://img.shields.io/badge/nextflow%20DSL2-%E2%89%A521.10.3-23aa62.svg?labelColor=000000)](https://www.nextflow.io/)
```
maxulysse, line 9: this one was out of sync with the minimal version.
Should there have been a linting failure?
cc @mirpedrol

.github/workflows/ci.yml
mirpedrol: The version is updated in the current dev version of tools, if the template was updated using v2.8 I think this is expected

modules/local/rename_raw_data_files.nf
```diff
@@ -5,7 +5,7 @@ process RENAME_RAW_DATA_FILES {
     conda "conda-forge::sed=4.7"
     container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
         'https://depot.galaxyproject.org/singularity/ubuntu:20.04' :
-        'docker.io/ubuntu:20.04' }"
+        'nf-core/ubuntu:20.04' }"
```
d4straub, line 8: The docker.io was specifically mentioned in https://nfcore.slack.com/archives/C04QR0T3G3H/p1682587842705369, why the change to nf-core?

adamrtalbot: Hmmm that advice is a bit wrong now - if you have a choice you should use quay.io hosted container. `quay.io/nf-core/ubuntu:20.04` is a container we control so better for long term maintenance than the `docker.io/library/ubuntu:20.04`.

.github/workflows/ci.yml
d4straub: nf-core/tools version 2.8 uses 22.10.1, see https://github.com/nf-core/tools/blob/0912990a63ef29e44e07cc2ba6ab81113684e0ae/nf_core/pipeline-template/nextflow.config#L225

README.md
```diff
@@ -6,7 +6,7 @@
...
+[![Nextflow](https://img.shields.io/badge/nextflow%20DSL2-%E2%89%A523.04.0-23aa62.svg?labelColor=000000)](https://www.nextflow.io/)
```
d4straub, line 9: nf-core/tools version 2.8 uses 22.10.1, see https://github.com/nf-core/tools/blob/0912990a63ef29e44e07cc2ba6ab81113684e0ae/nf_core/pipeline-template/nextflow.config#L225

d4straub: But indeed, 21.10.3 is wrong, should be at least 22.10.1 if above is true

.github/workflows/ci.yml
mirpedrol: The Nextflow version will change in v2.9, it's required in order to use `singularity.registry` https://github.com/nf-core/tools/pull/2305

conf/test_full.config
```diff
@@ -10,8 +10,6 @@
 ----------------------------------------------------------------------------------------
 */
 
-cleanup = true
```
d4straub, line 13: > This should go, otherwise we don't keep results

Thanks, but when/where do we not keep results? Because in nf-core/mag it seems to appear as well: https://github.com/nf-core/mag/blob/66cf53aff834d2a254b78b94fc54cd656b8b7b57/conf/test_full.config#L13 and it seems to keep results on the [website](https://nf-co.re/mag/results) (in case you mean that...).

modules/local/rename_raw_data_files.nf
maxulysse: Actually not really a need there, but we have this one on quay.io, and they're less picky than docker.io

nextflow.config
```diff
@@ -299,8 +304,8 @@ manifest {
     homePage        = 'https://github.com/nf-core/ampliseq'
     description     = """Amplicon sequencing analysis workflow using DADA2 and QIIME2"""
     mainScript      = 'main.nf'
-    nextflowVersion = '!>=22.10.1'
-    version = '2.6.0'
+    nextflowVersion = '!>=23.04.0'
```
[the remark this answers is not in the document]
maxulysse: to be able to use `singularity.registry`

conf/test_full.config
maxulysse: Let me double check that one

.github/workflows/ci.yml
d4straub: Alright, so we are adding here nf-core/tools dev changes instead of 2.8? Why? That isnt at all my goal.

nextflow.config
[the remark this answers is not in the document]
d4straub: And why would we need that?

modules/local/rename_raw_data_files.nf
d4straub: Thanks for the info!

conf/test_full.config
maxulysse: yeah, so not to prevent resume.
sorry, I was a bit ahead of myself

d4straub: Ah I see, alright, thanks!

README.md
d4straub: Ok, since its not rly 2.8 any more...

.github/workflows/ci.yml
d4straub: Well, whatever works.

d4straub (approved):
Thanks a lot, looks good, I hope AWS trouble is solved.
Only the CHANGELOG might need an update!

maxulysse:
sorry I forgot about the CHANGELOG

PR #616, opened 2023-08-08
Use nf-validation on samplesheet

d4straub:
nf-core template v2.9 came with nf-validation that can be also applied to validate samplesheets.

However, nf-validation cannot handle multiple input kinds, such as we have for `--input` (samplesheet, ASV/OTU fasta, folder input). Only with separate input params, nf-validation can be used for the sample sheet. Therefore I split `--input` into 
- `--input` (samplesheet), 
- `--input_fasta` (ASV/OTU fasta), and 
- `--input_folder` (folder input). 

I am unsure whether we need or want that. I can also add a few lines in `parse_input.nf` instead that will check at least for sampleID regex. Any opinions welcome.

This PR isnt complete, documentation must be further updated and code could be a little more cleaned (e.g. is_fasta_input wouldnt be needed any more). But I will only invest in this if we decide to actually do this.

`test_single` will fail, because the samplesheet contains sampleID that starts with numbers, we want to avoid that in future.

## PR checklist


erikrikarddaniel:
Fine to have three input params, I think. Actually clearer in many ways compared to the implicit handling of today.

workflows/ampliseq.nf
```diff
@@ -222,13 +215,46 @@ workflow AMPLISEQ {
...
+    // Create input channels
+    //
+    ch_input_fasta = Channel.empty()
+    ch_input_reads = Channel.empty()
+    if ( params.input ) {
+        // See the documentation https://nextflow-io.github.io/nf-validation/samplesheets/fromSamplesheet/
+        ch_input_reads = Channel.fromSamplesheet("input")
+            .map{ meta, readfw, readrv ->
+                meta.single_end = single_end.toBoolean()
+                def reads = single_end ? readfw : [readfw,readrv]
```
d4straub, line 227: Hi @erikrikarddaniel ,

all seems fine now, but I think that line here will cause a pipeline error when:
1. samplesheet contains only forward reads
2. `--single_end` isnt specified

before the PR here 2. wasnt needed. 
I think changing to
```suggestion
                def reads = single_end || readrv == null ? readfw : [readfw,readrv]
```
will make sample sheet input with only forward reads and without `--single_end` pass (need to test). Would that be desireable? That could lead in mixed entries though, such as some samples being treated as single ended and some as paired-end. That would crush later the pipeline I guess.

erikrikarddaniel: Haven't we required `--single_end` before and not relied on what the sample sheet looked like? (I thought so, but you seem to claim otherwise, and I suppose you're right.) I think it's fine to continue with that behaviour (or introduce it here, if required) if that simplifies handling of the input sheet.

In principle, it would be nice to be able to mix, but in most cases that would mean one wants to specify sample specific primer sequences too. That's maybe overkill?

erikrikarddaniel: Now I see edit2: Fine for me.

d4straub: Thanks for your feedback, I'll test how the pipeline behaves with single-end samplesheet without specifying --single_end and will add an error message if needed. Otherwise that PR is god to go (after review) I think.

> In principle, it would be nice to be able to mix, but in most cases that would mean one wants to specify sample specific primer sequences too. That's maybe overkill?

Yes theoretically that would be not bad. But that would require some refactoring, i.e. samplesheet with optional primer sequences, primers into meta map, and some trouble such as using primer sequences for taxonomic database that would need to be matched. Probably not very problematic, but for my projects not worth it. Abundance tables also can be always merged on ASV md5sums of multiple ampliseq runs or such. I think it will complicate the pipeline considerably more, but I that would need to be evaluated if someone wants to implement that ;)

CHANGELOG.md
```diff
@@ -11,6 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Changed`
 
+- [#616](https://github.com/nf-core/ampliseq/pull/616) - When using a sample sheet with `--input` containing forward an reverse reads, specifying `--single_end` will only extract forward reads and treat the data as single ended instead of extracting forward and reverse reads.
```
erikrikarddaniel, line 14:
```suggestion
- [#616](https://github.com/nf-core/ampliseq/pull/616) - When using a sample sheet with `--input` containing forward and reverse reads, specifying `--single_end` will only extract forward reads and treat the data as single ended instead of extracting forward and reverse reads.
```

docs/usage.md
```diff
@@ -94,16 +94,71 @@ You can also generate such `YAML`/`JSON` files via [nf-core/launch](https://nf-c
...
+
+- [Samplesheet input](#samplesheet-input) using `--input`: Tab-separated samplesheet
+- [ASV/OTU fasta input](#asvotu-fasta-input) using `--input_fasta`: Fasta file with sequences to be taxonomically classified
+- [Direct FASTQ input](#direct-fastq-input) using `--input_folder`: Folder containing zipped FastQ files.
 
 Optionally, a metadata sheet can be specified for downstream analysis.
 
+#### Samplesheet input
+
+The sample sheet file can be tab-separated (.tsv), comma-separated (.csv), or in YAML format (.yml/.yaml) that must have two to four columns/entries with the following headers:
```
erikrikarddaniel, line 108:
```suggestion
The sample sheet file can be tab-separated (.tsv), comma-separated (.csv), or in YAML format (.yml/.yaml) and can have two to four columns/entries with the following headers:
```

nextflow_schema.json
```diff
@@ -73,7 +88,8 @@
                 },
                 "single_end": {
                     "type": "boolean",
-                    "description": "If data is single-ended Illumina reads instead of paired-end"
+                    "description": "If data is single-ended Illumina reads instead of paired-end",
+                    "help_text": "When using a sample sheet with `--input` containing forward an reverse reads, specifying `--single_end` will only extract forward reads and treat the data as single ended instead of extracting forward and reverse reads."
```
erikrikarddaniel, line 92:
```suggestion
                    "help_text": "When using a sample sheet with `--input` containing forward and reverse reads, specifying `--single_end` will only extract forward reads and treat the data as single ended instead of extracting forward and reverse reads."
```

erikrikarddaniel (approved):
Looks great. I only found a few spelling mistakes.

workflows/ampliseq.nf
d4straub: > Haven't we required --single_end before and not relied on what the sample sheet looked like? 

Indeed, I checked, that was the case.

d4straub: I added a line below this one here making sure that when there isnt any reverse read there must be single_end=true.

d4straub:
Thanks for the review & advice!

PR #621, opened 2023-08-18
Update nextflow_schema.json

jtangrot:
Just a typo fix, it's probably more appropriate to write VSEARCH sintax (instead of USEARCH)


d4straub (approved):
LGTM

PR #637, opened 2023-09-13
Add taxonomic classification with kraken2

d4straub:
Kraken2 was reported to perform well on amplicon data, e.g. https://doi.org/10.1186/s40168-020-00900-2 & https://www.nature.com/articles/s41598-023-40799-x.
I decided here to go with the idea of the pipeline to re-create ASVs and therefore used Kraken2 on ASVs instead of on raw reads.
Kraken2 supports silva, rdp, and greengenes, but also whole genome databases. I added of the latter also the "standard" database which is huge but includes the typical microbes. Potentially the "standard" database would allow analyzing any amplicon, e.g. rarely used genes that have no proper gene specific database.
Additionally, a custom (i.e. any) Kraken2 database can be used, even if here not added in the `ref_databases.config`.
Added four parameter: `--kraken2_ref_taxonomy`, `--kraken2_ref_tax_custom`, `--kraken2_assign_taxlevels`, `--kraken2_confidence`

I looked into using confidence thresholds for taxonomic assignments, which are essentially k-mer match ratios for the LCA calculation, see https://github.com/DerrickWood/kraken2/blob/v2.1.3/docs/MANUAL.markdown#confidence-scoring. There is also an interesting discussion in `https://github.com/DerrickWood/kraken2/issues/265` about confidence scores. However, I have the impression that there isnt a consensus and all of this is referring to shotgun metagenomics. When looking at the amplicon sequencing paper (https://doi.org/10.1186/s40168-020-00900-2), I couldnt find any mention of the confidence score but bracken was used after kraken2. So for now I added a parameter for the confidence threshold but kept it by default at 0, which is also kraken2's default.
I am fine with when 1 k-mer causes e.g. a species instead of a genus annotation (i.e. a low confidence score for species), because I expect that only very rarely a k-mer is really specific. However, conflicting annotations (lets say 2 k-mer on genus a and 1 k-mer on genus b) would be good to capture (and report family in the example instead of genus a). Not sure if thats done right now as expected. 
There is a tool that calculates confidence scores for kraken2, see https://github.com/Ivarz/Conifer, which is also in [bioconda](https://anaconda.org/bioconda/conifer). Potentially interesting.

## PR checklist

- [x] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/ampliseq/tree/master/.github/CONTRIBUTING.md)
- [x] `README.md` is updated (including new tool citations and authors/contributors).


tests/pipeline/reftaxcustom.nf.test
```diff
@@ -41,6 +41,9 @@ nextflow_pipeline {
                 { assert new File("$outputDir/fastqc/sampleID_2a_1_fastqc.html").exists() },
                 { assert new File("$outputDir/fastqc/sampleID_2a_2_fastqc.html").exists() },
                 { assert snapshot(path("$outputDir/input/Samplesheet.tsv")).match("input") },
+                { assert snapshot(path("$outputDir/kraken2/ASV_tax.user.kraken2.classifiedreads.txt"),
+                                path("$outputDir/kraken2/ASV_tax.user.kraken2.complete.tsv"),
+                                path("$outputDir/kraken2/ASV_tax.user.kraken2.tsv")).match("barrnap") },
```
d4straub, line 46:
```suggestion
                                path("$outputDir/kraken2/ASV_tax.user.kraken2.tsv")).match("kraken2") },
```

tests/pipeline/doubleprimers.nf.test
```diff
@@ -30,6 +30,9 @@ nextflow_pipeline {
                                 path("$outputDir/dada2/DADA2_table.rds"),
                                 path("$outputDir/dada2/DADA2_table.tsv")).match("dada2") },
                 { assert snapshot(path("$outputDir/input/Samplesheet_double_primer.tsv")).match("input") },
+                { assert snapshot(path("$outputDir/kraken2/ASV_tax.greengenes.kraken2.classifiedreads.txt"),
+                                path("$outputDir/kraken2/ASV_tax.greengenes.kraken2.complete.tsv"),
+                                path("$outputDir/kraken2/ASV_tax.greengenes.kraken2.tsv")).match("barrnap") },
```
d4straub, line 35:
```suggestion
                                path("$outputDir/kraken2/ASV_tax.greengenes.kraken2.tsv")).match("kraken2") },
```

d4straub (commented):
Still typos...

conf/ref_databases.config
```diff
@@ -382,4 +382,74 @@ params {
             dbversion = "UNITE-alleuk v8.2 (https://dx.doi.org/10.15156/BIO/786376)"
         }
     }
+    // Kraken2 reference databases
+    kraken2_ref_databases {
+        // Corresponding S3 URLs can be obtained by removing https://genome-idx.s3.amazonaws.com from the beginning of the URLs linked to above and replacing with s3://genome-idx.
+        'silva' {
+            title = "Kraken2 pre-formatted SILVA - Version 138"
+            file = [ "https://genome-idx.s3.amazonaws.com/kraken/16S_Silva138_20200326.tgz" ]
```
FriederikeHanssen, line 390: are the DBs versioned or can the data behind the link change at any time?

modules/local/format_taxresults_kraken2.nf
```diff
@@ -0,0 +1,107 @@
+process FORMAT_TAXRESULTS_KRAKEN2 {
+    label 'process_low'
+
+    conda "bioconda::bioconductor-dada2=1.22.0 conda-forge::r-digest=0.6.30"
+    container "${ workflow.containerEngine == 'singularity' && !task.ext.singularity_pull_docker_container ?
+        'https://depot.galaxyproject.org/singularity/bioconductor-dada2:1.22.0--r41h399db7b_0' :
+        'biocontainers/bioconductor-dada2:1.22.0--r41h399db7b_0' }" // TODO: pure R seems sufficient!
```
FriederikeHanssen, line 7: what about this todo?

FriederikeHanssen (approved):
looks good, I can't say anything regarding the biological side of things

conf/ref_databases.config
d4straub: The key 'silva' has always the latest version, currently its identical with 'silva=138', 'silva=132' is an older version. The files chosen when using 'silva' can indeed change from release to release.

FriederikeHanssen: so no reproducibility 😢 ?

d4straub: Given the same pipeline version, 'silva' will always be the same version. Across pipeline versions, 'silva' will be always the latest version, i.e. differ. By using 'silva=138' the database will be fixed across pipeline versions.
I would call that reproducible.

modules/local/format_taxresults_kraken2.nf
d4straub: Good catch! Forgot about it. I check if I can use R only.

conf/ref_databases.config
FriederikeHanssen: ah ok, then all good, forget my comment :) I misunderstood you earlier, I thought you meant it updates to latest on every run.

modules/local/format_taxresults_kraken2.nf
d4straub: R-only was successful, changed it, thanks for spotting it!

d4straub:
Thanks for your review!

PR #638, opened 2023-09-19
Catch error case where 0 reads pass filtering and params.ignore_failed_trimming = True

sminot:
In my use of this workflow I found an edge case where the filtering step produces 0 reads. In those cases when `params.ignore_failed_trimming = True`, those samples should be ignored. However, instead I observed that the workflow failed with:

```
Missing output file(s) `*.filt.fastq.gz` expected by process NFCORE_AMPLISEQ:AMPLISEQ:DADA2_PREPROCESSING:DADA2_FILTNTRIM
```

By adding the `optional` flag to the output using `params.ignore_failed_trimming`, this error will be avoided only in those cases when it satisfies the request of the user.

My apologies for not adding tests, but this is such a small change that I thought it might be acceptable.

## PR checklist

- [ ] Make sure your code lints (`nf-core lint`) (this failed, but not in any part of the code that I touched)


d4straub:
Hi there,

thanks for providing a fix! I havent come across that case yet.
However, I have concerns with this PR. 
- the change was added to the master branch, but except the CHANGELOG conflicts that might work anyway.
- params.ignore_failed_trimming is supposed to be a param for cutadapt, e.g. when after cutadapt the number of reads fall below a threshold (--min_read_counts, default: 0), but I agree the naming is ambiguous so that could be also used here
- there isnt any notification when a sample fails the threshold and `params.ignore_failed_trimming = true`, it just continues silently, right? That is a little concerning but probably also fine.
- if that error occurs, as you have described above, there is no indication that this could be solved by using --ignore_failed_trimming, correct? I think thats maybe the most severe problem. When the problem occurs its not clear that there is a solution at hand.

So the optimal solution to me seems at that point that (1) there is a hint when an error occurs at that step to use a specific parameter, probably mentioning the samples that failed, (2) when the parameter is used a warning lists samples that failed but the pipeline continues. Essentially how https://nf-co.re/ampliseq/2.6.1/parameters/#ignore_empty_input_files & https://nf-co.re/ampliseq/2.6.1/parameters/#ignore_failed_trimming work. My idea would be to create empty output files when there is none produced by the filtntrim command and then filter the outgoing channel with nextflow's `.branch`. Would that make sense?

d4straub:
As a note, this is easily reproducible by running `nextflow run nf-core/ampliseq -r 2.6.1 -profile test,singularity --outdir results  --trunclenf 300 --trunclenr 300` where trunclenf/r will cause discarding all reads because its PE250.

sminot:
I really appreciate your taking a look at this PR, @d4straub , and I must apologize for not putting it together as thoroughly as I should have. The nf-core/ampliseq workflow is really amazingly helpful, and I appreciate all the work that you all have put into it.

I'm trying to come up with a test case that recapitulates my experience in which 0 reads are output by the `DADA2_FILTNTRIM` process, such that no files are created and the process ends with an error. I'm getting the test suite all staged and running on my local system, and I'll see if I can find the right kind of inputs which sufficiently resemble this particular (admittedly extremely low quality) dataset.

I like your idea of using empty outputs from the process, since that would ensure that the dropped samples are still logged as being missing in the same way that the current skipping behavior will document. Given the utility of this workflow, I think it will be worthwhile to add that particular feature.

I'll work on getting that running using a new test dataset which is appropriate for this edge case. I'll update you when I have something pushed that is ready for your inspection.

d4straub:
Thanks! I was testing a little bit and actually, the solution of https://nf-co.re/ampliseq/2.6.1/parameters/#ignore_empty_input_files & https://nf-co.re/ampliseq/2.6.1/parameters/#ignore_failed_trimming, i.e. having a verbose error, having an empty file and filtering it with `.branch`, reporting samples and warnings, went fine until it clashed in:
- https://github.com/nf-core/ampliseq/blob/b602d1908ac7d43c71624526990c7c43a8ade45f/workflows/ampliseq.nf#L343-L347
where it couldnt `.join`, because sample ids, which are in `meta.id`, didnt match
- after removing `meta.id`, the pipeline failed in a subsequent step where meta.id was required (dont have my notes with me, not sure it was)
at that point I found it got too complicated. So probably that approach is not perfect in that case. But I definately didnt come up with an elegant solution.
Probably you do it better, just wanted to share my failed attempt that you might benefit from it ;)

sminot:
This is actually really relevant @d4straub, since I'm trying to debug a dataset which stops at basically the exact point you describe, but which I thought should still pass filtering for at least some samples.

I'm doing more local testing, and I'll see if I can figure this one out and propose a good fix.

sminot:
Closing in favor of https://github.com/nf-core/ampliseq/pull/641

PR #641, opened 2023-09-26
Continue analysis even when individual files fail the filtering threshold

sminot:
This PR addresses the issue that when samples fail to pass the filtering threshold, they will:

1. Raise an error in DADA2_FILTANDTRIM when no files are written out, and
2. Cause no outputs to be written when the id element fails to match on future join steps

I've added a test case to reproduce the bug, and the code I've added should address it.

This should be a better solution than what is outlined in #638 (which I'm now closing)

## PR checklist


sminot:
Test data in https://github.com/nf-core/test-datasets/pull/1008

d4straub:
Thanks! That looks indeed elegant, going to test it today and give feedback.

subworkflows/local/dada2_preprocessing.nf
```diff
@@ -74,20 +74,52 @@ workflow DADA2_PREPROCESSING {
...
+                log.warn "The following samples had too few reads (<$params.min_read_counts) after quality filtering with DADA2:\n$samples\nIgnoring failed samples and continue!\n"
+            } else {
+                error("The following samples had too few reads (<$params.min_read_counts) after quality filtering with DADA2:\n$samples\nPlease check whether the correct primer sequences for trimming were supplied. Ignore that samples using `--ignore_failed_filtering` or adjust the threshold with `--min_read_counts`.")
+            }
+        }
+
+    // Break apart the reads and logs so that only the samples
+    // which pass filtering are retained
+    ch_dada2_filtntrim_results_passed
+        .map{ it -> [it[0], it[1]] }
```
d4straub, line 100: That seems more verbose, would it work? If yes the other two might benefit as well, I like more verbose
```suggestion
        .map{ meta, reads, logs, args -> [meta, reads] }
```

conf/test_failed.config
d4straub, line 1: For me, the test fails in
`NFCORE_AMPLISEQ:AMPLISEQ:QIIME2_ANCOM:QIIME2_ANCOM_TAX`
with 
```
All values in `grouping` are unique. This method cannot operate on a grouping vector with only unique values (e.g., there are no 'within' variance because each group of samples contains only a single sample).
```
and the problem are
`materialSampleID`
`bad9`
etc
This is because the automatic choosing of the metadata columns fails, because now some columns that had before unique values (should be omitted) are included and after removal of the sample unique again. Overwriting it with e.g. `--metadata_category "treatment1"` makes the test pass.
Another option is to just remove the unnecessary columns in the sample sheet.

nextflow.config
```diff
@@ -263,6 +264,7 @@ profiles {
     test_pacbio_its    { includeConfig 'conf/test_pacbio_its.config'    }
     test_iontorrent    { includeConfig 'conf/test_iontorrent.config'    }
     test_fasta         { includeConfig 'conf/test_fasta.config'         }
+    test_failed        { includeConfig 'conf/test_failed.config'        }
```
d4straub, line 267: Will need an entry in `.github/workflows/ci.yml` as well.

d4straub (commented):
Awesome, thanks that was a good idea to change the ouput of `dada2_filtntrim` in that way!

The metadata however makes the `test_failed` fail for me, easily fixable though.

Changelog could also get an update, you can add yourself to the contributors in the readme credits section as well if you want.

conf/test_failed.config
sminot: Great idea, I should just trim it down

sminot:
Ok, I think I'm in good shape now, @d4straub 

- Test data is merged
- Added to CI
- Added to CHANGELOG
- Made the channel filtering code more explicit as you suggested

Let me know what you think!

d4straub (approved):
Great thanks, all looks good.

However, there is one more problem (apologies for not having predicted this earlier, that is a relatively new feature). For the new way of testing with github, at least a "failed.nf.test" in `tests/pipeline/` seems required. That file specifies tests that output files are present, and can in conjunction with "failed.nf.test.snap" even check for md5sums. I think the presence/md5sum is only needed for central files, here probably 
cutadapt/cutadapt_summary.tsv
barrnap/summary.tsv
dada2/DADA2_table.tsv
overall_summary.tsv
But I can also add that when I do the template update, so I approve anyway. However, it would be nice if you want to still add that last piece!

sminot:
The test validation piece is not something I'm familiar with, so I might just vote for merging as-is and adding in those additional files in a subsequent PR.

It does appear that I cannot merge with the tests failing. If you could squash and merge that would be greatly appreciated!

Thanks also for all your help as I was getting this together @d4straub. My first real contribution to nf-core!

PR #644, opened 2023-10-12
Update pipeline summary report

d4straub:
Ironing out issues in the pipeline summary report, custom taxonomic databases weren't properly reported (NA).
Additionally moved a code snippet from `ampliseq.nf` to a subworkflow, a purely cosmetic change.

## PR checklist

### DIFF
--- a/assets/report_template.Rmd
+++ b/assets/report_template.Rmd
@@ -1692,8 +1692,12 @@ cat(filter_codons_fasta_passed,"ASVs had no stop codons (",params$stop_codons,")
 
 ```{r, eval = any_taxonomy, results='asis'}
 methods_taxonomic_classification <- c("Taxonomic classification was performed by ")
-if ( !isFALSE(params$dada2_ref_tax_title) ) {
-    methods_taxonomic_classification_dada <- paste("DADA2 and the database '",params$dada2_ref_tax_title,"' (`", params$dada2_ref_tax_citation,"`)", sep="")
+if ( !isFALSE(params$dada2_taxonomy) ) {
+    if ( !isFALSE(params$dada2_ref_tax_title) ) {
+        methods_taxonomic_classification_dada <- paste("DADA2 and the database '",params$dada2_ref_tax_title,"' (`", params$dada2_ref_tax_citation,"`)", sep="")
+    } else {
+        methods_taxonomic_classification_dada <- paste("DADA2 with a user provided database", sep="")
+    }
     if ( !isFALSE(params$cut_dada_ref_taxonomy) ) {
         methods_taxonomic_classification_dada <- paste(methods_taxonomic_classification_dada,
             "that had",cut_dada_ref_taxonomy_filt,"sequences extracted by PCR primers to improve assignments")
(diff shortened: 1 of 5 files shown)


assets/report_template.Rmd
```diff
@@ -894,7 +892,7 @@ invisible(dev.off())
 cat("## DADA2\n")
 
 # indicate reference taxonomy
-if (!params$flag_dada_ref_tax_user) {
+if (params$dada2_ref_tax_title) {
```
d4straub, line 895:
```suggestion
if ( !isFALSE(params$dada2_ref_tax_title) ) {
```

assets/report_template.Rmd
```diff
@@ -1126,7 +1106,7 @@ cat("\n\nSINTAX taxonomy assignments can be found in folder [sintax](../sintax).
 cat("## Kraken2\n")
 
 # indicate reference taxonomy
-if (!params$flag_kraken2_ref_tax_user) {
+if (params$kraken2_ref_tax_title) {
```
d4straub, line 1109:
```suggestion
if (  !isFALSE(params$kraken2_ref_tax_title) ) {
```

assets/report_template.Rmd
```diff
@@ -894,7 +892,7 @@ invisible(dev.off())
 cat("## DADA2\n")
 
 # indicate reference taxonomy
-if (!params$flag_dada_ref_tax_user) {
+if ( !isFALSE(params$dada2_ref_tax_title) ) {
```
erikrikarddaniel, line 895: This looks odd to me -- negating `isFALSE` instead of testing for true -- but maybe it's the way it should be?
(In (at least) one more place further down.)

subworkflows/local/dada2_preprocessing.nf
```diff
@@ -136,7 +136,8 @@ workflow DADA2_PREPROCESSING {
         ch_DADA2_QUALITY2_SVG = DADA2_QUALITY2.out.svg
     }
 
-    //group by sequencing run
+    //group reads by sequencing run
+    //for 'groupTuple', 'size' or 'groupKey' should be used but to produce it we need to know how many elements to group but some can be lost here, so no way knowing before
```
erikrikarddaniel, line 140:
```suggestion
    // 'groupTuple', 'size' or 'groupKey' should be used but to produce it we need to know how many elements to group but some can be lost here, so no way knowing before
```

erikrikarddaniel (approved):
Difficult to review due to the complexity, but since tests pass it must be fine.

Two tiny comments.

assets/report_template.Rmd
d4straub: True that looks odd. I actually havent tested `isTRUE` but used `!isFALSE` because the default of that params is `FALSE`, if data is coming through it could be anything, TRUE, a string (file, setting, ...), a number,...
So I figured `!isFALSE` is my safest bet. Am I wrong?

erikrikarddaniel: I've never used either of the two functions, but `isTRUE` would be my choice even if the default for the param is false. The difference between `if ( isTRUE(cond) )` and `if ( cond )` appears to be that `isTRUE` doesn't return NAs. `isFALSE` is defined analogously according to the documentation. `isTRUE('string')` returns false while evaluating a string as a logical returns NA. If you're not certain of what type the variable is (e.g. a logical or a string) I think you need to do something more elaborate.

d4straub: Yes, the point is here that I'm not certain of what type the variable is (e.g. a logical or a string) and 
```
!isFALSE('string') 
!isFALSE(TRUE)
!isFALSE(1)
```
all return TRUE, so seem to work as intended.
`if ( 'string' ) {} `returns `Error in if ("string") { : argument is not interpretable as logical`
`isTRUE('string')` is `FALSE`, so not doing what I want.
So while I agree the double negation looks bad, it seems to do what it should and the cleaner versions I came up with didnt.

d4straub: I am open to suggestion ofc ;)

erikrikarddaniel: I suppose the correct thing would be to make sure the variable is logical before using it. I understand if you have better things to do than correct working code though.

PR #667, opened 2023-11-28
Added QIIME2 custom reference database support.

MatthewJM96:
Added support for using custom reference databases in QIIME2 taxonomic classification via the `--qiime_ref_tax_custom` flag. This brings QIIME2 taxonomic classification into alignment with Kraken and Dada which allow the same.

Testing should probably be added, I could do with some advice on how to make this possible with some reduced database that matches the requirement on what can be passed to the flag (must be a directory or tarball as in the Kraken implementation).

## PR checklist


nextflow_schema.json
```diff
@@ -432,6 +432,11 @@
                         "greengenes85"
                     ]
                 },
+                "qiime_ref_tax_custom": {
```
d4straub, line 435: Usually `--*_ref_tax_custom` comes with a `--*_assign_taxlevels`, is that not needed here?

subworkflows/local/qiime2_preptax.nf
```diff
@@ -2,20 +2,58 @@
...
 workflow QIIME2_PREPTAX {
     take:
     ch_qiime_ref_taxonomy //channel, list of files
+    val_qiime_ref_taxonomy //val
     FW_primer //val
     RV_primer //val
 
     main:
-    FORMAT_TAXONOMY_QIIME ( ch_qiime_ref_taxonomy )
+    if (params.qiime_ref_tax_custom) {
```
d4straub, line 18: Why did you choose a tarball instead of e.g. of a comma separated list of compressed files, such as `--qiime2_ref_tax_custom "seq.fna.gz,taxonomy.tax.gz"`? I think the tarball has the disadvantage of requiring un-compressing before checking that files with required regex are present while the list of files can be checked with nextflow code immediately. Is there somewhere a resource for tar'ed QIIME2 databases as for Kraken2?

workflows/ampliseq.nf
d4straub, line 1: Would be nice to forward the classification results also to the markdown pipeline summary report in https://github.com/nf-core/ampliseq/blob/1067c7c2c88861635955f32adc7a2682885fe27a/workflows/ampliseq.nf#L842

d4straub (commented):
Thanks for that PR!
Looks good to me, see comments below.

I think a proper test file could be created with greengenes85 with files https://github.com/nf-core/ampliseq/blob/1067c7c2c88861635955f32adc7a2682885fe27a/conf/ref_databases.config#L305 and uploaded to https://github.com/nf-core/test-datasets/tree/ampliseq/testdata (maybe into a new folder such as "DB") if small enough and activated in e.g. https://github.com/nf-core/ampliseq/blob/dev/conf/test_reftaxcustom.config which would require an update of https://github.com/nf-core/ampliseq/blob/dev/tests/pipeline/reftaxcustom.nf.test and https://github.com/nf-core/ampliseq/blob/1067c7c2c88861635955f32adc7a2682885fe27a/tests/pipeline/reftaxcustom.nf.test.snap#L16

lib/WorkflowAmpliseq.groovy
```diff
@@ -77,20 +77,20 @@ class WorkflowAmpliseq {
...
             Nextflow.error("Incompatible parameters: `--FW_primer` and `--RV_primer` are required for cutting the QIIME2 reference database to the amplicon sequences. Please specify primers or do not use `--qiime_ref_taxonomy`.")
         }
 
         if ( (!params.FW_primer || !params.RV_primer) && params.cut_dada_ref_taxonomy && !params.skip_taxonomy ) {
             Nextflow.error("Incompatible parameters: `--FW_primer` and `--RV_primer` are required for cutting the DADA2 reference database to the amplicon sequences. Please specify primers or do not use `--cut_dada_ref_taxonomy`.")
         }
 
-        if (params.qiime_ref_taxonomy && params.classifier) {
+        if ((params.qiime_ref_taxonomy || params.qiime_ref_tax_custom) && params.classifier) {
             Nextflow.error("Incompatible parameters: `--qiime_ref_taxonomy` will produce a classifier but `--classifier` points to a precomputed classifier, therefore, only use one of those.")
```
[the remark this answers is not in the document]
MatthewJM96: Added this change in with a small adjustment.

subworkflows/local/qiime2_preptax.nf
MatthewJM96: I've added support for a pair of comma separated files. I saw that the unite databases were pulled as a single tarball, and figure users may prepare databases that they want to move around easier as a single file.

I've not added gzip support for files when pointing to a directory to give some better consistency, but I wanted to get some feedback first on if the flexibility of supporting different input forms is desired.

workflows/ampliseq.nf
MatthewJM96: Updated this and a couple of other downstream flow that needed updating.

subworkflows/local/qiime2_preptax.nf
```diff
@@ -2,20 +2,81 @@
...
+        // Handle case where we have been provided a pair of filepaths.
+        if ("${params.qiime_ref_tax_custom}".contains(",")) {
+            ch_qiime_ref_taxonomy.flatten()
+                .branch {
+                    compressed: it.isFile() && it.getName().endsWith(".gz")
+                    decompressed: it.isFile() && ( it.getName().endsWith(".fna") || it.getName().endsWith(".tax") )
+                    failed: true
+                }.set { ch_qiime_ref_tax_branched }
+            ch_qiime_ref_tax_branched.failed.subscribe { error "$it is neither a compressed or decompressed sequence or taxonomy file. Please review input." }
```
d4straub, line 30:
```suggestion
            ch_qiime_ref_tax_branched.failed.subscribe { error "$it is neither a compressed (ends with `.gz`) or decompressed sequence (ends with `.fna`) or taxonomy file (ends with `.tax`). Please review input." }
```
I think thats more transparent otherwise someone might be surprised their "*.fasta" isnt a decrompressed fasta file ;)

subworkflows/local/qiime2_preptax.nf
d4straub: Great, I think thats all one can ask for.
It would be optimal to cover as many inputs as possible in tests configs.

MatthewJM96:
> Thanks for that PR! Looks good to me, see comments below.
> 
> I think a proper test file could be created with greengenes85 with files
> 
> https://github.com/nf-core/ampliseq/blob/1067c7c2c88861635955f32adc7a2682885fe27a/conf/ref_databases.config#L305
> and uploaded to https://github.com/nf-core/test-datasets/tree/ampliseq/testdata (maybe into a new folder such as "DB") if small enough and activated in e.g. https://github.com/nf-core/ampliseq/blob/dev/conf/test_reftaxcustom.config which would require an update of https://github.com/nf-core/ampliseq/blob/dev/tests/pipeline/reftaxcustom.nf.test and
> 
> https://github.com/nf-core/ampliseq/blob/1067c7c2c88861635955f32adc7a2682885fe27a/tests/pipeline/reftaxcustom.nf.test.snap#L16

I've begun putting some testing in, I guess based on supporting a few input forms it might be good to create a couple of smaller tests like reftaxcustom just for qiime2. Especially as right now reftaxcustom wants to disable downstream analysis after dada and kraken2. I could similarly disable some downstream things for qiime but right now that's controlled by run_qiime2.

In our usecase, this work is a predecessor to adding consensus blast processing in qiime2, which I'll try to get in working order for a PR in time. Does it therefore make sense to separate out the logic that sets run_qiime2 for differentiating between the downstream analysis in qiime and the taxonomic alignment in qiime?

erikrikarddaniel:
[...]
> > Does it therefore make sense to separate out the logic that sets run_qiime2 for differentiating between the downstream analysis in qiime and the taxonomic alignment in qiime?
> 
> `run_qiime2` is used for taxonomic classification [here](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/workflows/ampliseq.nf#L553-L572) and for downstream analysis [here](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/workflows/ampliseq.nf#L574-L577). I guess it could make sense to separate those (maybe [here](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/workflows/ampliseq.nf#L134)) into `run_qiime2_downstreamanaylsis` and `run_qiime2_taxonomy` and potentially add another one for blast consensus (or keep blast consensus & scikit learn in "taxonomy"). Just do it as intuitive as possible and as easy to maintain as possible (keep checks to a minimum).

The idea would be that some users might want QIIME's taxonomy but not all the rest? If so, why not keep `--skip_qiime` but allow it to be combined with `--qiime_taxonomy` instead of the quite long params you suggest?

MatthewJM96:
> > I've begun putting some testing in
> 
> Yes, separate tests are fine if its not really possible to fit into existing tests.

I've put a test with tarball into the existing reftaxcustom case, and added a qiimecustom that tests with a file pair. I think that's a reasonable balance to test different input patterns.

> > Does it therefore make sense to separate out the logic that sets run_qiime2 for differentiating between the downstream analysis in qiime and the taxonomic alignment in qiime?
> 
> `run_qiime2` is used for taxonomic classification [here](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/workflows/ampliseq.nf#L553-L572) and for downstream analysis [here](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/workflows/ampliseq.nf#L574-L577). I guess it could make sense to separate those (maybe [here](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/workflows/ampliseq.nf#L134)) into `run_qiime2_downstreamanaylsis` and `run_qiime2_taxonomy` and potentially add another one for blast consensus (or keep blast consensus & scikit learn in "taxonomy"). Just do it as intuitive as possible and as easy to maintain as possible (keep checks to a minimum).

> The idea would be that some users might want QIIME's taxonomy but not all the rest? If so, why not keep `--skip_qiime` but allow it to be combined with `--qiime_taxonomy` instead of the quite long params you suggest?

I've added a `--skip_qiime_downstream` flag and separated out calculation of a `run_qiime2` that applies to downstream and a `run_qiime2_taxonomy` that applies just to the taxonomy stage, using this is the test cases.

nextflow_schema.json
```diff
@@ -651,6 +656,10 @@
                     "type": "boolean",
                     "description": "Skip all steps that are executed by QIIME2, including QIIME2 software download, taxonomy assignment by QIIME2, barplots, relative abundance tables, diversity analysis, differential abundance testing."
                 },
+                "skip_qiime_downstream": {
+                    "type": "boolean",
+                    "description": "Skip steps that are executed by QIIME2 except for taxonomic classification, including barplots, relative abundance tables, diversity analysis, differential abundance testing."
```
d4straub, line 661:
```suggestion
                    "description": "Skip steps that are executed by QIIME2 except for taxonomic classification. Skip steps including barplots, relative abundance tables, diversity analysis, differential abundance testing."
```
I am obviously not a native speaker, but this was its clearer to me, does that fit?

d4straub (commented):
Hi Matthew,

looks great. Did you run your newly added tests and `test` itself and made sure files look fine? If not I'll have a look before I give my ok here.

MatthewJM96:
> Hi Matthew,
> 
> looks great. Did you run your newly added tests and `test` itself and made sure files look fine? If not I'll have a look before I give my ok here.

I ran them manually but also added the new test to the CI pipeline and it looks like it passed. I couldn't figure a good way to snapshot the QIIME taxonomic classification as I guess the algorithm isn't deterministic, so I just checked that the classifier and taxonomy tsv report are produced.

nextflow_schema.json
MatthewJM96: Sorry I didn't reply to this yet. We haven't made use of these flags in our team. If I am understanding the effects of the `--*_assign_taxlevels` flags right, qiime2 doesn't need an equivalent. Qiime2 doesn't take such an argument as Dada does, and it produces a report compatible with downstream analysis by definition so reports wouldn't need to be adapted as the Kraken and Sintax ones are using the same flag.

If I'm mistaken, happy to chat about how to make use of such a flag.

d4straub:
Thanks, I have the feeling I should also run a few tests just to make sure, I scheduled some time tomorrow, so I expect to approve the PR then.

d4straub:
I tested, and found:
(1) there is something wrong with phyloseq, I attempted to fix it in https://github.com/nf-core/ampliseq/pull/676
(2) when running `nextflow run MatthewJM96/ampliseq -r qiime2_custom_db -profile test_qiimecustom,singularity --outdir result_test_qiimecustom_qiime2_custom_db_23-12-12` I found that `result_test_qiimecustom_qiime2_custom_db_23-12-12/summary_report/summary_report.html` didnt contain the section about the taxonomy, https://github.com/nf-core/ampliseq/pull/673 should fix that.

I think the ideal sequence should be:
(1) After https://github.com/nf-core/ampliseq/pull/676 is merged
(2) integrate dev into that PR
(3) revert https://github.com/nf-core/ampliseq/pull/667/commits/4464c38cef7be3e9309c3d036fda7172aba130a4
(4) all should be fine (check if `summary_report.html` is fine with `-profile test_qiimecustom`) and we merge.

Sorry that you run into that phyloseq bug, I hope it works now.

d4straub:
I found the problem and fixed the report. When all tests passed I'll merge it if you do not have any objections.

d4straub (approved):
LGTM!

MatthewJM96:
> I found the problem and fixed the report. When all tests passed I'll merge it if you do not have any objections.

Sorry for no reply, was on leave! Thanks for looking at those last things and the advice along the way.

PR #672, opened 2023-12-06
Update output docs for collapsed abundance tables

danclaytondev:
Hi,

This PR is a minor output documentation update.

Recently I was confused why the qiime output relative abundance tsvs only went to level 6, not level 7 when the reference database we are using includes level 7. In the parameter schema, it shows that the level is controlled by `--tax_agglom_max` which defaults to 6, so hopefully the output docs better reflect the situation with this change.

I agree that level 6 is a sensible default, I am not proposing to change that, just the output docs! :)

To verify the changed documentation, I believe this is the code that collapses the taxa for [absolute abundance counts](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/modules/local/qiime2_export_absolute.nf#L57-L75), and [relative abundance counts](https://github.com/nf-core/ampliseq/blob/113e90bdff42a52807f5c8b3cbaafaa31c145b9d/modules/local/qiime2_export_reltax.nf#L30-L52).

@d4straub Apologies, reopened the PR after fixing the committer details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### `Changed`
 
 ### `Fixed`
+
 - [#672](https://github.com/nf-core/ampliseq/pull/672) - Update output docs for collapsed abundance tables
 
 ### `Dependencies`


d4straub (approved):
Looks good to me!
Do you want to merge it (after all tests below pass)?

Issue #670, opened 2023-12-04
Add option to assign ASV to multiple species with DADA2 


andand:
### Description of feature

Corresponding to setting `allowMultiple = TRUE` in `assignSpecies` or `addSpecies`.

erikrikarddaniel:
Would make sense. I can foresee some parsing problems of the resulting output file, but nothing that can't be dealt with. A test case would hence be good.

erikrikarddaniel:
I just realized you can probably do this right away by creating a `nextflow.config` file with the following content:

```
process {
    withName: DADA2_ADDSPECIES {
        ext.args = "allowMultiple = TRUE, n = 1e5"
    }
}
```

If this breaks the output or not, I'm not sure, but if you try please report what happened.

d4straub:
Added that option to dev, will be in release 2.8.0.

PR #681, opened 2023-12-19
Add two parameters for taxonomic classification with DADA2

d4straub:
1. Added two parameters:

- `--dada_addspecies_allowmultiple`: Defines the behavior when multiple exact matches against different species are returned. By default only unambiguous identifications are returned. If TRUE, a concatenated string of all exactly matched species is returned.
- `--dada_taxonomy_rc`: Reverse-complement of each sequences will be used for classification if it is a better match to the reference sequences than the forward sequence.

3. Updated `usage.md `to explain that `--dada_taxonomy_rc` is recommended with `--dada_ref_taxonomy gtdb` according to https://zenodo.org/records/4735821:

> Another concern that was raised was the orientation of the DB sequences, to get past this problem please use the tryRC = TRUE argument in the assignTaxonomy command within DADA2, this will search your ASVs in the reverse complement as well. 

4. Updated the default of `--kraken2_confidence` because it was always showing up as changed even if it was default.

Closes https://github.com/nf-core/ampliseq/issues/670

## PR checklist


d4straub:
Hi @erikrikarddaniel , I added here a parameter to allow multiple exact matches, I could not see anywhere trouble. However, could there be any consequences to SBDI?

erikrikarddaniel (approved):
:+1: 
Since the default is to keep the old behaviour, I can't see any problem with this.

d4straub:
Thanks!

Issue #704, opened 2024-02-11
ampliseq fails during taxonomy assignation when processing ITS sequences 


eperezv:
### Description of the bug

Hello,
I was processing my ITS data using ampliseq and I found a problem. The log can be found below. 

I think the problem is directly related to the option
>--cut_its

I have checked many things and I think the problem is caused by ITSX, which produces a few very short ASVs (<50 nt) that crash R during taxonomic assignation. I don't know why ITSx is producing that very short ASVs (I think it shouldn't, or there should be a way of filtering them out). I have tried to remove short ASVs with the --min_len_asv but it seems it works at other step and not right before taxonomic assignation. 

My solution for now run ampliseq, wait for it to crash. Then, manually remove short ASVs and re-run ampliseq.

I also have a feature suggestion, which is allowing the user to choose the taxonomic group in ITSx. I modified the code manually because I am only interested in amplicons related to fungi.

This is the log:  

```
Feb-11 18:49:48.111 [Task submitter] DEBUG nextflow.processor.TaskProcessor - Handling unexpected condition for
  task: name=NFCORE_AMPLISEQ:AMPLISEQ:DADA2_TAXONOMY_WF:DADA2_TAXONOMY (ASV_ITS_seqs.ITS2.fasta,assignTaxonomy.fna); work-dir=/home/user/ampliseq/work>
  error [nextflow.exception.ProcessUnrecoverableException]: Process requirement exceeds available memory -- req: 128 GB; avail: 125.6 GB
Feb-11 18:49:48.120 [Task submitter] ERROR nextflow.processor.TaskProcessor - Error executing process > 'NFCORE_AMPLISEQ:AMPLISEQ:DADA2_TAXONOMY_WF:DADA2_TAXONOMY (ASV_ITS_seqs.ITS2.fasta,assignTaxonomy.>

Caused by:
  Process requirement exceeds available memory -- req: 128 GB; avail: 125.6 GB

Command executed:

  #!/usr/bin/env Rscript
      suppressPackageStartupMessages(library(dada2))
      set.seed(100) # Initialize random number generator for reproducibility

      taxlevels <- c("Kingdom", "Phylum", "Class", "Order", "Family", "Genus", "Species")

      seq <- getSequences("ASV_ITS_seqs.ITS2.fasta", collapse = TRUE, silence = FALSE)
[...]

PR #718, opened 2024-03-25
Require a minimum sequence length of 50bp after ITSx

d4straub:
Sequences below 50bp cannot be used by DADA2 and most likely also other kmer-based classification methods fail with too short sequences (see `https://github.com/benjjneb/dada2/issues/601`). This can be usually easily prevented with `--min_len_asv`, but not when ITSx is used (because this is after `--min_len_asv` is appplied). Therefore, another length filtering step with 50bp min length is added here. The minimum value of 50 is not represented by a parameter but can be adjusted via a config, if required.

Closes https://github.com/nf-core/ampliseq/issues/704.

## PR checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#711](https://github.com/nf-core/ampliseq/pull/711) - From r207 and onwards Archaea sequences were omitted when parsing GTDB databases. (This did not affect `sbdi-gtdb` databases, only `gtdb`.)
 - [#715](https://github.com/nf-core/ampliseq/pull/715) - Fix filtering vsearch clusters for high number of clusters
 - [#717](https://github.com/nf-core/ampliseq/pull/717) - Fix edge case for sorting file names by using radix method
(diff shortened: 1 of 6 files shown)


erikrikarddaniel (approved):
:+1:

d4straub:
Thanks!

jtangrot (approved):
Looks good! Thanks! My only concern is that the length filtering is generally applied, and not only when running DADA2 for taxonomies. E.g. sintax by default discards sequences shorter than 32, not 50, but this can be changed with a parameter and I think there is no requirement on sequence length (not going into details on how short a sequence can be to achieve sensible results). So for sintax a sequence length of 50 is not technically needed. That being said, I guess it's easy to change the filtering cutoff in ampliseq with a custom config if one thinks 50 is too much.

d4straub:
Thanks for your review! Unfortunately I seem to have merged the PR just before your comment.
I actually do not know what methods require a minimal length and which not, just DADA2 seems certain.
I considered 2 solutions: (1) adding a filter after ITSx to control its output for all downstream processes, or (2) adding a length filter before DADA2.
My impression was that short sequences are generally not advisable to classify (and didnt know that sintax has a build-in filter) and therefore I decided for option (1). Would you have preferred option (2)?

jtangrot:
Yes, I realized I was too slow when posting my comment :D I'm fine with this solution too.

PR #719, opened 2024-03-26
Export all process software versions

d4straub:
Until now only software versions of selected processes were listed, this is changed here so that software versions of all processes are now exported to `pipeline_info/software_versions.yml`.

## PR checklist

### DIFF
--- a/modules/local/format_taxonomy.nf
+++ b/modules/local/format_taxonomy.nf
@@ -31,7 +31,6 @@ process FORMAT_TAXONOMY {
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        bash: \$(bash --version | sed -n 1p | sed 's/GNU bash, version //g')
-    END_VERSIONS
+        bash: \$(bash --version | sed -n 1p | grep -Eo 'version [[:alnum:].]+' | sed 's/version //')
     """
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Changed`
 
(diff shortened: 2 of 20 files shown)


modules/local/format_taxonomy.nf
```diff
@@ -31,7 +31,6 @@ process FORMAT_TAXONOMY {
 
     cat <<-END_VERSIONS > versions.yml
     "${task.process}":
-        bash: \$(bash --version | sed -n 1p | sed 's/GNU bash, version //g')
-    END_VERSIONS
+        bash: \$(bash --version | sed -n 1p | grep -Eo 'version [[:alnum:].]+' | sed 's/version //')
```
erikrikarddaniel, line 34:
```suggestion
        bash: \$(bash --version | sed -n 1p | grep -Eo 'version [[:alnum:].]+' | sed 's/version //')
        END_VERSIONS
```
Doesn't one need this? (I see you removed it in several places.)

erikrikarddaniel (approved):
Looks good, except for some missing `END_VERSIONS` -- maybe it works anyway?

modules/local/format_taxonomy.nf
d4straub: Yes that is really weird: The `END_VERSIONS` was ending up in the `versions.yml` and the pipeline failed. So I removed it and the tests passed. I am not sure why. If there is a better solution, that would be great indeed, any idea?

erikrikarddaniel: That sounds like it's oddly placed in the generated script. It should be to right to the left without whitespace, I think.

erikrikarddaniel: I see that my suggestion will actually place whitespaces before `END_VERSIONS`, so don't accept that!

d4straub: Thanks, I'll have a look

d4straub: Indeed, there was weird indentation after variables which hold scripts from bin. We'll see if thats resolved now.

Issue #811, opened 2024-12-09
AMPLISEQ:RENAME_RAW_DATA_FILES pipeline break


Dedaniya08:
### Description of the bug

The pipeline break when rename the raw data files, when running the AMPLISEQ: RENAME_RAW_DATA_FILES. in Azure Computer instance.

### Command used and terminal output

```console
$ nextflow run nf-core/ampliseq -r 2.11.0 -profile docker -c ./final_run_folder/nextflow.config,./final_run_folder/conf/test_modified.config, -params-file params.yaml --multiple_sequencing_runs -w ./working_dirr/
```

### Relevant files

[error.txt](https://github.com/user-attachments/files/18057990/error.txt)

### System information

Azure Computer Instance:

* Nextflow version ( 24.10.0.5928)
* Standard_F64s_V2 (64 cores, 128 GB RAM, 512 GB disk)
* Executor (local)
* Container engine: Docker
* OS (debain ubuntu)
* nf-core/ampliseq (2.11.0)

erikrikarddaniel:
In the error file it says `Cannot run program "/bin/bash"` which sounds critical to me, but I don't know a thing about Azure computing. I think your problem relates to how you set up your run and is not a bug in the pipeline. I would therefore recommend you to post your issue at the nf-core Slack: https://nf-co.re/join/slack, since there are people there that knows more about cloud computing in general as well as Azure in particular.

d4straub:
Closing this because that wont be solved here unfortunately, partially also discussed in #813

Issue #812, opened 2024-12-09
AMPLISEQ:DADA2_PREPROCESSING: DADA2_FILTNTRIM pipeline break


Dedaniya08:
### Description of the bug

The pipeline break when the output files is missed in fetching *.filt.fastq.gz' which is expected by the process 'NFCORE_AMPLISEQ:AMPLISEQ:DADA2_PREPROCESSING:DADA2_FILTNTRIM'

The pipeline break and not wait till the process executed.

### Command used and terminal output

```console
$ nextflow run nf-core/ampliseq -r 2.11.0 -profile docker -c ./final_run_folder/nextflow.config,./final_run_folder/conf/test_modified.config, -params-file params.yaml --multiple_sequencing_runs -w ./working_dirr/
```

### Relevant files

![Image](https://github.com/user-attachments/assets/6290d7f6-c426-4118-9c12-69d8cc3a5e87)

### System information

Azure Computer Instance:

Nextflow version ( 24.10.0.5928)
Standard_F64s_V2 (64 cores, 128 GB RAM, 512 GB disk)
Executor (local)
Container engine: Docker
OS (debain ubuntu)
nf-core/ampliseq (2.11.0

d4straub:
Closing this, explained in #813

PR #813, opened 2024-12-09
 Fixed the issue #811 and #812, which leading the pipeline break

Dedaniya08:
## PR checklist


Dedaniya08:
The issue of pipeline breaking is caused at two steps for issue #811 and #812 . I have added the retry max 3 attempt which has lead to resolving the both the error in both the files
second_step


Both the step are resolved.

d4straub:
Hi there, thanks for the idea. However, this is a very bad idea to retry for any error code. This can lead to a lot of confusion. I oppose that change.
We typically only retry on specific error codes, see https://github.com/nf-core/ampliseq/blob/8f139cecd95a23920617f2c2b154733494ed4562/conf/base.config#L18.

If that change solves you problem thats great, but there is no need to change pipeline code instead you can achieve this also by using `-c azure.conf` that contains:
```
process {
    errorStrategy = 'retry'
}
```
or use more process targeted approach as in e.g. https://github.com/nf-core/ampliseq/blob/8f139cecd95a23920617f2c2b154733494ed4562/conf/base.config#L65-L69

Dedaniya08:
The code is encountering frequent issues. In this PR, step #811 focuses on renaming the files, and step #812 maps the filtered files. I verified the files after a retry, and the output remains consistent. There are no discrepancies in the files before and after applying this PR.

The recurring issue is due to Azure's local environment and the process of extracting files from the data lake. This behavior will persist without the changes introduced here. Therefore, I strongly recommend including this PR to address the problem effectively.

d4straub:
Thanks for explaining your reasoning. I do believe that the resulting files are fine and I acknowledge that this proposed change will solve the issue on Azure. But it will negatively affect the execution of the pipeline on any other platform (retry for all exit codes of `RENAME_RAW_DATA_FILES` & `DADA2_FILTNTRIM`, which is not desirable). I would prefer a fix that doesn't have a negative impact at all.
I'll try to lure some more people here to get more opinions.

modules/local/dada2_filtntrim.nf
```diff
@@ -1,6 +1,8 @@
 process DADA2_FILTNTRIM {
     tag "$meta.id"
     label 'process_medium'
+    errorStrategy 'retry'
+    maxRetries 3
```
ewels, line 5: This is [already set](https://github.com/nf-core/ampliseq/blob/8f139cecd95a23920617f2c2b154733494ed4562/conf/base.config#L19) globally at `process` level, so no need to set again.
```suggestion
```

modules/local/rename_raw_data_files.nf
```diff
@@ -1,6 +1,8 @@
 process RENAME_RAW_DATA_FILES {
     tag "$meta.id"
     label 'process_low'
+    errorStrategy 'retry'
+    maxRetries 3
```
ewels, line 5:
```suggestion
```

ewels (changes requested):
This issue seems related to the Azure infrastructure setup and is not general to all users of the pipeline, or really pipeline specific. I do not think that we should apply blanket `retry` for all users like this.

Alternatives would be setting this on user configs instead, or better still adding to the [`azurebatch` institutional config](https://nf-co.re/configs/azurebatch/) as a [pipeline-specific config](https://github.com/nf-core/configs?tab=readme-ov-file#adding-a-new-pipeline-specific-config).

The latter is already set up in this pipeline so no changes needed here, purely nf-core/configs repo. Need to create a new config file there and then include it [here](https://github.com/nf-core/configs/blob/master/pipeline/ampliseq.config).

ewels:
It's also possible that the `retry` solves the issue here because the retry attempts give the task more memory (the `task.attempt` multiplier). This could be needed when pulling data from a data lake, which is a slightly non-standard method of data staging.

https://github.com/nf-core/ampliseq/blob/8f139cecd95a23920617f2c2b154733494ed4562/conf/base.config#L30

So something else that you could try in the azure-specific pipeline config is to assign more memory to these tasks. Then they might not fail in the first place and you may not need the `retry`.

ewels:
@d4straub - suggest we close this PR and @Dedaniya08 you can open one to nf-core/configs with the relevant changes.

d4straub:
Yes I agree @ewels , better solved in nf-core/configs. Thanks a lot!

Issue #857, opened 2025-03-25
ITSx filters out ASVs that remain in the table causing the diversity core to crash


tom-brekke:
### Description of the bug

Hello, 
Following up from slack discussion here: https://nfcore.slack.com/archives/CEA7TBJGJ/p1740756972411109

There seems to be a bug when using ITSx in conjunction with the QIIME2 diversity core. In the attached example ITSx claims to find ITS in all 35 ASVs (see ASV_ITS_seqs.summary.txt), but then only writes a subset (31) of them to the output file called AVS_ITS_seq.full.fasta. The QIIME2 tree is built using the smaller fasta resulting in a tree with 31 tips. The error arises when this tree is compared to the table which has the original 35 ASVs during the QIIME2_DIVERSITY_CORE step. 

Perhaps the easiest solution for this bug is to not use the ITSx output fasta for tree-building?  

[ASV_ITS_seqs.summary.txt](https://github.com/user-attachments/files/19449743/ASV_ITS_seqs.summary.txt)
[ASV_ITS_seqs.full.fasta.txt](https://github.com/user-attachments/files/19449742/ASV_ITS_seqs.full.fasta.txt)
[ASV_seqs.len.fasta.txt](https://github.com/user-attachments/files/19449776/ASV_seqs.len.fasta.txt)

### Command used and terminal output

```console
nextflow run nf-core/ampliseq -r 2.12.0 -profile apptainer -resume -params-file nf-params_PacBioFungalOTU_UNITE_2_4kreads.json -c qiime2-2024-10.config

 N E X T F L O W   ~  version 24.10.5

Launching `https://github.com/nf-core/ampliseq` [nostalgic_picasso] DSL2 - revision: 8f139cecd9 [2.12.0]


------------------------------------------------------
                                        ,--./,-.
        ___     __   __   __   ___     /,-._.--~'
  |\ | |__  __ /  ` /  \ |__) |__         }  {
  | \| |       \__, \__/ |  \ |___     \`-._,-`-,
                                        `._,._,'
  nf-core/ampliseq 2.12.0
------------------------------------------------------
Main arguments
  input                              : /home/tbrekke/scratch/private/Fungal_OTU_PacBio_NCEA_8/03_ampliseq/ampliseq_samplesheet_2_4kreads.tsv
  FW_primer                          : CCTCCSCTTANTDATATGC
  RV_primer                          : CCTNGTTGATYCTGCCAGT
[...]

PR #862, opened 2025-03-26
Fix ITSx filtering

d4straub:
Closes https://github.com/nf-core/ampliseq/issues/857

ITSx removed some sequences without explicitly stating so, but the abundance table remained untouched. Now the abundance table is filtered based on the sequences that pass ITSx.

## PR checklist


modules/local/filter_len.nf
nschan, line 1: This seems good, but I don't think that using `seq` and `list` (which are both base R functions) as names for objects is a great idea.

d4straub: Indeed, that seems even a quite bad idea. I'll see if I can quickly rename.

PR #863, opened 2025-03-26
Add bold-trnl configuration

georgiakes:
Added the bold-trnl database.
Still missing the addignSpecies reference file and the script for formating.


d4straub:
Great!
- we sorted the entries roughly alphabetically to keep easier overview, so please put it more to the top ;)
- its mandatory to state some sort of version, since the db must be versioned
- please use two entries: (1) without version, i.e. `bold-trnl`, (2) with version, i.e. `bold-trnl=<version>`. (1) always points to the most recent db version. examples in the file.

Issue #827, opened 2025-01-24
Run classification step in chunks in case of large number of ASVs


jtangrot:
### Description of feature

As discussed in slack, the classification step (in particular DADA2_ADDSPECIES) use excessive amounts of RAM in cases when there are lots of ASVs generated, e.g. in the order of 60-100 000 ASVs. A suggested solution is to run the classification step in batches by splitting the ASV_seqs.fasta file, and then merge the resulting taxonomy files. It would be great if it would be possible to implemented that in the pipeline instead of having to run it manually.

d4straub:
Thanks!
Chunking can be done as described in https://nextflow-io.github.io/patterns/process-per-file-chunk/ and files can be collected afterwards and merged again. Should not be too complicated to implement. One parameter for the chunk size might be good so that one can split to the desired size and have by default probably 10k or so.
[unfortunately I dont see currently a time window where I could do that]

jtangrot:
Unfortunately I have no time to spend on this myself at the moment

d4straub:
Seems not too complicated, I try to finish ASV chunking for DADA2 in the last minutes of the hackathon.

PR #864, opened 2025-03-26
Classify sequences with DADA2 in chunks

d4straub:
Closes https://github.com/nf-core/ampliseq/issues/827

Still to do:
- documentation
- changelog

## PR checklist

### DIFF
--- a/nextflow.config
+++ b/nextflow.config
@@ -126,6 +126,7 @@ params {
     cut_dada_ref_taxonomy    = false
     dada_addspecies_allowmultiple = false
     dada_taxonomy_rc         = false
+    dada_assign_chunksize    = 10000
     sintax_ref_taxonomy      = null
     qiime_ref_taxonomy       = null
     qiime_ref_tax_custom     = null
--- a/nextflow_schema.json
+++ b/nextflow_schema.json
@@ -483,6 +483,12 @@
                     "help_text": "Reverse-complement of each sequences will be used for classification if it is a better match to the reference sequences than the forward sequence.",
                     "description": "If reverse-complement of each sequences will be also tested for classification"
                 },
+                "dada_assign_chunksize": {
+                    "type": "integer",
+                    "help_text": "Chunks for DADA2's assignTaxonomy and addSpecies can speed up the process and lower required memory.",
+                    "description": "ASV fasta will be subsetted into chunks of that size for classification",
+                    "default": 10000
+                },
                 "pplace_tree": {
                     "type": "string",
                     "description": "Newick file with reference phylogenetic tree. Requires also `--pplace_aln` and `--pplace_model`."
--- a/conf/modules.config
+++ b/conf/modules.config
@@ -519,6 +519,11 @@ process {
                 path: { "${params.outdir}/dada2/args" },
                 mode: params.publish_dir_mode,
                 pattern: "*.args.txt"
+            ],
+            [
+                path: { "${params.outdir}/dada2/chunks" },
+                mode: params.publish_dir_mode,
+                pattern: "*.tsv"
             ]
         ]
     }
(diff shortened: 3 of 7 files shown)


d4straub:
I broke something: phyloseq & tse are only processing _either_ dada2 _or_ qiime2 taxonomic assignments with profile `test`.

modules/local/phyloseq.nf
```diff
@@ -8,9 +8,7 @@ process PHYLOSEQ {
         'biocontainers/bioconductor-phyloseq:1.46.0--r43hdfd78af_0' }"
 
     input:
-    tuple val(prefix), path(tax_tsv), path(otu_tsv)
-    path sam_tsv
-    path tree
+    tuple val(prefix), path(tax_tsv), path(otu_tsv), path(sam_tsv), path(tree)
```
erikrikarddaniel, line 11: This is interesting. It's been a point of mine for a while for nf-core modules that it's often better to have a single multi-entry parameter as it allows to keep the arguments synced. But I saw the other day that that's not following the recommendations. Always nice to be proven right! ;-)

nextflow.config
```diff
@@ -126,6 +126,7 @@ params {
     cut_dada_ref_taxonomy    = false
     dada_addspecies_allowmultiple = false
     dada_taxonomy_rc         = false
+    dada_assign_chunksize    = 10000
```
erikrikarddaniel, line 129: Isn't that a bit low? If one wants to keep the original behaviour one could set this to something ridiculously high, e.g. 1000000.

nextflow_schema.json
```diff
@@ -483,6 +483,12 @@
                     "help_text": "Reverse-complement of each sequences will be used for classification if it is a better match to the reference sequences than the forward sequence.",
                     "description": "If reverse-complement of each sequences will be also tested for classification"
                 },
+                "dada_assign_chunksize": {
+                    "type": "integer",
+                    "help_text": "Chunks for DADA2's assignTaxonomy and addSpecies can speed up the process and lower required memory.",
+                    "description": "ASV fasta will be subsetted into chunks of that size for classification",
```
erikrikarddaniel, line 489:
```suggestion
                    "description": "ASV fasta will be subset into chunks of this size for classification",
```

erikrikarddaniel (approved):
LGTM, with a few comments.

nextflow.config
d4straub: I actually wanted specifically to have a realistic value so that chunking goes into effect when it might be beneficial (I didnt test when it is, might depend whether the compute resource is large enough to run all chunks in parallel). Do you suggest its better to keep the original behavior by choosing a high default? I am unsure whats best here.

erikrikarddaniel: I have no strong opinion, and maybe you're right. We've had a couple of complaints about this, so maybe keep as it is now.

docs/output.md
```diff
@@ -272,6 +272,7 @@ Files when _not_ using ITSx (default):
   - `ASV_tax.*.tsv`: Taxonomic classification for each ASV sequence.
   - `ASV_tax_species.*.tsv`: Exact species classification for each ASV sequence.
   - `ref_taxonomy.*.txt`: Information about the used reference taxonomy, such as title, version, citation.
+- `dada2/chunks/`: Directory containing files with taxonomy assignments of individual chunks.
```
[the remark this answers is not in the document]
d4straub: I checked, no that seems to follow the standard in this file

modules/local/phyloseq.nf
d4straub: I actually preferred the previous input line, but I wasnt able to get it running stable without changing it, especially because both, sam_tsv and tree, are optional inputs.

nextflow_schema.json
d4straub: Thanks!

PR #907, opened 2025-08-29
Fix split_regions error with only single sample

4less:
Single line change in Rscript.
With only one sample available the fixed line would drop data.frame format and rowSums throws an error. with "drop=FALSE" the data.frame format is retained. 

## PR checklist


4less:
Accidentally included other commits

PR #913, opened 2025-09-25
Add min boot param

zmudge3:
## PR checklist

This PR adds a new parameter, dada_min_boot, that corresponds exactly with minBoot in DADA2's [assignTaxonomy](https://rdrr.io/bioc/dada2/man/assignTaxonomy.html) method. This parameter specifies a minimum bootstrap confidence (out of 100 trials) needed in order to assign a taxonomic level. This can be helpful in order to filter out assignments for ASVs that are likely too specific. E.g. An ASV was assigned to the incorrect species with confidence 55 (0.55), which is quite low, but was assigned to the correct genus level with confidence 75 (0.75). By setting dada_min_boot to 60 (0.60) in this example, the reported assignment would be to the genus level. Users can adjust dada_min_boot based on their own observations concerning what threshold they begin to see inaccurate assignments.

This also contains a small commit to adjust the order of the read truncation parameters (truncq, trunc_qmin, trunc_rmin, trunclenf, trunclenr) in nextflow.config to match the order they appear in nextflow_schema.json. No functional difference.

### DIFF
--- a/nextflow_schema.json
+++ b/nextflow_schema.json
@@ -405,6 +405,23 @@
             },
             "fa_icon": "fas fa-filter"
         },
+        "dada2_assigntaxonomy": {
+            "title": "DADA2 assignTaxonomy",
+            "type": "object",
+            "description": "Options for the DADA2 assignTaxonomy method used to classify ASVs against a database",
+            "default": "",
+            "fa_icon": "fas fa-sitemap",
+            "properties": {
+                "dada_min_boot": {
+                    "type": "integer",
+                    "default": 50,
+                    "fa_icon": "fas fa-greater-than-equal",
+                    "description": "The minimum bootstrap confidence (out of 100 trials) for assigning a taxonomic level. Matches `minBoot` in DADA2's assignTaxonomy method.",
+                    "minimum": 0,
+                    "maximum": 100
+                }
+            }
+        },
         "taxonomic_database": {
             "title": "Taxonomic database",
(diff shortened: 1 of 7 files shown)


nextflow_schema.json
```diff
@@ -405,6 +405,23 @@
             },
             "fa_icon": "fas fa-filter"
         },
+        "dada2_assigntaxonomy": {
```
d4straub, line 408: You added here a new section here. I have some trouble with that:
- dada_taxonomy_rc also is directly related to "DADA2 assignTaxonomy" and should accordingly appear in this section here
- dada_assign_taxlevels seems also to fit into "DADA2 assignTaxonomy"
- the headline "DADA2 assignTaxonomy" is very technical (too technical for a headline imho), less technical would be e.g. "Taxonomic classification with DADA2", that would include then options for assignTaxonomy & addSpecies & potentially also cut_dada_ref_taxonomy, dada_assign_chunksize,  etc; however, that would then imply also separate sections for kraken2 and such.

I would prefer if you do not put dada_min_boot into a new section but rather into the section "Taxonomic database". I agree that this header is too narrow for all the params that are listed there. We could rename it to something like "Taxonomic assignment" or "Taxonomic classification" (I think there are a few links in the docs to that header/section, so that links would need to updated).

d4straub (changes requested):
All fine to me, except the parameter documentation.

PR #949, opened 2026-02-12
Add metadata filter

d4straub:
This is adding the ability to filter the data for downstream analysis with the metadata sheet. This could be helpful to remove negative control samples meant for Decontam or subset the data for e.g. PCoA plot and the like. As a side effect, when samples are not passing preprocessing (and that is intentionally ignored by setting specific parameters) then the pipeline will not complain that the samples are in the metadata sheet but not in the other data tables.

CI test `test_failed` (3/14) was modified via https://github.com/nf-core/test-datasets/commit/311e901643905c9c24de55aadf0768dc1ea9826b to test for that case.

Due to the change from 4 to 3 samples to downstream analysis, beta diversity statistics analysis cannot be performed, leading to less files produced. Additionally, the count tables change their content (and therefore md5sum).

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/ampliseq/tree/master/.github/CONTRIBUTING.md)
- [x] If necessary, also make a PR on the nf-core/ampliseq _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [ ] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

PR #959, opened 2026-03-11
Phylogenetic placement in multiple trees with HMM profiles

erikrikarddaniel:
## PR checklist

Finally, I've added the new mode for phylogenetic placement in which ASVs can be placed in one of several references.


erikrikarddaniel:
@nf-core-bot fix linting

erikrikarddaniel:
@nf-core-bot fix linting

assets/schema_pplace_sheet.json
```diff
@@ -0,0 +1,61 @@
...
+    "description": "Schema for the file provided with params.phylosearch_input",
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "target": {
+                "type": "string",
+                "pattern": "^\\S+$",
+                "errorMessage": "Target name must be provided and cannot contain spaces",
+                "meta": ["id"]
```
d4straub, line 14: I know `meta.id` is usually used, but in `schema_input.json` we use `meta.sample` from the samplesheet and fill `meta.id` later. Not sure (yet?) whether that is relevant, just as a comment for now.

conf/modules.config
```diff
@@ -759,18 +767,27 @@ process {
...
+
     withName: 'EPANG_PLACE' {
         ext.args   = { "--model ${meta.model}" }
         publishDir = [
             path: { "${params.outdir}/pplace/${task.process.tokenize(':')[-1].tokenize('_')[0].toLowerCase()}" },
             mode: params.publish_dir_mode,
             saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
         ]
+        cpus   = { 20    * task.attempt }
+        memory = { 60.GB * (1.5 ** (task.attempt - 1)) }
```
d4straub: I think any and all resource settings should be rather in `conf/base.config`. Its easier when all this is at one place, imho.

conf/modules.config
```diff
@@ -801,6 +818,8 @@ process {
                 saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
             ]
         ]
+        cpus   = { 20    * task.attempt }
+        memory = { 60.GB * (2 ** (task.attempt - 1)) }
```
d4straub: Same here, better into `conf/base.config`

docs/usage.md
```diff
@@ -340,9 +345,48 @@ The columns which are to be assessed can be specified by `--metadata_category`.
...
+archaea16s,clustalo,https://raw.githubusercontent.com/tseemann/barrnap/master/db/arc.hmm,16S_rRNA,archaea.newick,archaea.alnfna,GTR+F+I+G4,archaea.taxonomy.tsv
+bacteria16s,clustalo,https://raw.githubusercontent.com/tseemann/barrnap/master/db/bac.hmm,16S_rRNA,bacteria.newick,bacteria.alnfna,GTR+F+I+G4,bacteria.taxonomy.tsv
+euk18s,clustalo,https://raw.githubusercontent.com/tseemann/barrnap/master/db/euk.hmm,18S_rRNA,,,,
+```
+
+#### Placement in database-provided phylogenies
+
+Reference databases for taxonomy can provide reference phylogenies for placement (currently only SBDI-GTDB from release R10-RS226-2).
+This is not run by default since it takes quite a lot of resources.
+If you want to enable this, set `--skip_pplace false`.
```
d4straub: Personally I dont like that `skip_pplace` is set to `true` by default. I would rather like that the default is `pplace = false` and activate with `--pplace` (which is `false` by default) or such.

modules/local/format_taxonomy.nf
```diff
@@ -24,7 +24,6 @@ process FORMAT_TAXONOMY {
     echo -e "--dada_ref_taxonomy: ${params.dada_ref_taxonomy}\\n" >ref_taxonomy.${suffix}.txt
     echo -e "Title: ${params.dada_ref_databases[params.dada_ref_taxonomy]["title"]}\\n" >>ref_taxonomy.${suffix}.txt
     echo -e "Citation: ${params.dada_ref_databases[params.dada_ref_taxonomy]["citation"]}\\n" >>ref_taxonomy.${suffix}.txt
-    echo "All entries: ${params.dada_ref_databases[params.dada_ref_taxonomy]}" >>ref_taxonomy.${suffix}.txt
```
d4straub, line 27: Why did you remove that? Because it looks sort of ugly not with the additional of the pplace info? If thats the only reson, then I would like to keep it. Better ugly looking data than less data ;)

nextflow.config
```diff
@@ -116,6 +117,7 @@ params {
     skip_taxonomy          = false
     skip_dada_taxonomy     = false
     skip_dada_addspecies   = false
+    skip_pplace            = true
```
d4straub: I'd like a param with false as default, if possible at all.

d4straub (commented):
Very exciting addition!!!
First skimming looked good. I have a few comments. But it is close to the finish line I think.

conf/modules.config
erikrikarddaniel: I don't think I changed this. At least not intentionally. I was surprised to see it here. I'll have a look.

docs/usage.md
erikrikarddaniel: I think there's some recommendation someplace that we should rather use `--skip_*` than the other way around. I've seen that #mag has `--run_*`. I don't have a strong opinion myself, so I'll do what you recommend.

modules/local/format_taxonomy.nf
erikrikarddaniel: IIRC, something broke because of the very complex map we have now. I'll have another look.

assets/schema_pplace_sheet.json
erikrikarddaniel: I don't know if this matters at all. The key thing is that the user provides a column called `target`. However, since this is not a sample, but a name for the reference tree, I think we should not use `sample`.

conf/modules.config
erikrikarddaniel: I just removed this in the hope that the standard labels work.

assets/schema_pplace_sheet.json
d4straub: Agreed!

modules/local/format_taxonomy.nf
erikrikarddaniel: Works fine as it was, so I took it back.

erikrikarddaniel:
I'm opening this for review now. The `decontaminated_details.tsv` is unstable and I would remove it from md5sum tests. (It contains a lot of floating point numbers.) I tried making this more reproducible by sorting the table before writing it, but that didn't help.

.nf-core.yml
```diff
@@ -3,6 +3,10 @@ lint:
     - conf/igenomes.config
   files_unchanged:
     - .gitattributes
+    - assets/nf-core-ampliseq_logo_light.png
+    - assets/nf-core-ampliseq_logo_dark.png
+    - docs/images/nf-core-ampliseq_logo_light.png
+    - docs/images/nf-core-ampliseq_logo_dark.png
```
d4straub, line 9: Ah makes sense to avoid that annoyance.

nextflow_schema.json
```diff
@@ -952,6 +962,20 @@
...
         },
+        "running_specific_steps": {
+            "title": "Running specific steps",
+            "type": "object",
+            "description": "",
+            "default": "",
+            "properties": {
+                "run_pplace": {
+                    "type": "boolean",
+                    "description": "Run phylogenetic placement when reference trees are specified by the selected `--dada_ref_taxonomy`",
```
d4straub: Does that mean use pplace instead of dada2 or in addition to dada2? If its the latter, pplace taxonomies takes precedence over dada2 taxonomies for downstream analysis, right?

nextflow_schema.json
```diff
@@ -952,6 +962,20 @@
             },
             "fa_icon": "fas fa-hand-paper"
         },
+        "running_specific_steps": {
```
d4straub: I would probably put that section higher to make it more prominent. Probably even as the second, but definitely before skipping steps.

workflows/ampliseq.nf
```diff
@@ -693,14 +745,71 @@ workflow AMPLISEQ {
...
+    PPLACEFORMATTAX_SHEET(PPLACE_SHEET.out.taxonomy_per_query)
+    ch_versions = ch_versions.mix(PPLACEFORMATTAX_SHEET.out.versions)
+
+    // Currently, this is not used for anything since the QIIME_INTAX call is wrapped in an if cluase checking params.pplace_tree
+    if ( ! params.pplace_tree ) {
+        ch_pplace_tax = PPLACEFORMATTAX_SHEET.out.tsv
+            .splitCsv(sep: '\t', header: true)
+            .map { r -> "${r.ASV_ID}\t${r.taxonomy}\n" }
+            .collectFile(name: 'concatenated_taxonomy.tsv', seed: "ASV_ID\ttaxonomy\n")
+    }
```
d4straub: Not sure I understand this. What is not used? And could it be used by the downstream steps?

workflows/ampliseq.nf
```diff
@@ -679,7 +729,9 @@ workflow AMPLISEQ {
         ch_sintax_tax = channel.empty()
     }
 
-    // Phylo placement
+    // Phylogenetic placement
```
d4straub, line 759: Somehow it would be nice to better mark the whole new phylogenetic placement section as such. Now its nicely highlighted in the diff, but that wont be as easy to identify when the PR is merged. This particular headline looks even visually less important than the 
```
//
// SUBWORKFLOW: Run phylogenetic placement
//
```
below, but its at least the same importance, or even higher?

Possibly using here a subworkflow would make sense, but then the pplace subworkflow would contain other subworkflows? That seems also a little over-complicated.

d4straub (commented):
Looks quite good! Have some points still though.

> The decontaminated_details.tsv is unstable and I would remove it from md5sum tests. (It contains a lot of floating point numbers.) I tried making this more reproducible by sorting the table before writing it, but that didn't help.

That is surprising to me, because it seemed stable before, at least for the default.nf.test.snap even without sorting. Usually there are at least two tests running (minimum nf version & latest)to see if its just a new checksum or if its indeed unstable, but because of the trouble below I cannot check this. But if its indeed unstable, then omit the md5sum check.

The "latest-everything" tests fail with
```
    > Error modules/local/hmmer/hmmextract/main.nf:25:30: `keyfile` is not defined
    > │  25 |     def outfile = ! key && ! keyfile ? '' : "> ${prefix}.hmm"
    > ╰     |                              ^^^^^^^
    > 
    > 
    > ERROR ~ Script compilation failed
```
that might be worthwhile to fix because the upcoming nextflow releases might not work with that syntax as is. This is also described [here](https://github.com/nf-core/strict-syntax-health?tab=readme-ov-file#running-locally). But sometimes edge version have bugs...

nextflow_schema.json
erikrikarddaniel: Phyloplacement is run in addition to dada2, but no, I didn't implement this as the standard taxonomy for downstream purposes. One main reason is that not all ASVs will be placed if there are entries in the sheet without corresponding trees. (One can have entries with an hmm file that doesn't point to a tree. This is used to draw away false positive ASVs from placement in an unsuitable tree.) Another reason -- that we could circumvent -- is that in many cases there will be several trees and they might not have the same taxonomy (quite a hypothetical example, I admit).

workflows/ampliseq.nf
erikrikarddaniel: This is also because I found it difficult to motivate the use of these taxonomies for the reasons I give above. (It was also, IIRC, quite difficult to get right.)
I've updated the comment a bit since the channel is actually used in the call to `QIIME2_EXPORT`. Maybe this is actually misleading -- would it be better to leave this out of QIIME2 altogether?

erikrikarddaniel:
I'm opening this for review now. The `decontaminated_details.tsv` is unstable and I would remove it from md5sum tests. (It contains a lot of floating point numbers.) I tried making this more reproducible by sorting the table before writing it, but that didn't help.

> Looks quite good! Have some points still though.
> 
> > The decontaminated_details.tsv is unstable and I would remove it from md5sum tests. (It contains a lot of floating point numbers.) I tried making this more reproducible by sorting the table before writing it, but that didn't help.
> 
> That is surprising to me, because it seemed stable before, at least for the default.nf.test.snap even without sorting. Usually there are at least two tests running (minimum nf version & latest)to see if its just a new checksum or if its indeed unstable, but because of the trouble below I cannot check this. But if its indeed unstable, then omit the md5sum check.
> 
It's not stable for me. I've updated the test so it tests the md5sum of all decontam files except `decontaminated_details.tsv`. For this, I introduced a test of the number of lines in the file and that the `freq` column sums to 1.

> The "latest-everything" tests fail with
> 
> ```
>     > Error modules/local/hmmer/hmmextract/main.nf:25:30: `keyfile` is not defined
>     > │  25 |     def outfile = ! key && ! keyfile ? '' : "> ${prefix}.hmm"
>     > ╰     |                              ^^^^^^^
>     > 
>     > 
>     > ERROR ~ Script compilation failed
> ```
> 
> that might be worthwhile to fix because the upcoming nextflow releases might not work with that syntax as is. This is also described [here](https://github.com/nf-core/strict-syntax-health?tab=readme-ov-file#running-locally). But sometimes edge version have bugs...

Good that you saw this, I hadn't. (I will need to update #phyloplace too.)

workflows/ampliseq.nf
erikrikarddaniel: I marked it better in the code now.

nextflow_schema.json
erikrikarddaniel: Done.

modules/local/format_taxonomy.nf
d4straub: But its still removed?

docs/usage.md
d4straub: The docs here mention still `--skip_pplace false`?

nextflow_schema.json
d4straub: > One main reason is that not all ASVs will be placed if there are entries in the sheet without corresponding trees. (One can have entries with an hmm file that doesn't point to a tree. This is used to draw away false positive ASVs from placement in an unsuitable tree.)

But wouldnt it be theoretically fine to remove then any ASVs without taxonomic annotation from downstream analysis, because it was intentionally removed with a hmm without corresponding tree (decoy hmm)?
Or, alternatively, just forward all ASVs, even without taxonomic classification, into downstream analysis, because not all downstream steps need taxonomies?
Or, if both options above could be interesting, let the user decide with a param what option is happening?

> Another reason -- that we could circumvent -- is that in many cases there will be several trees and they might not have the same taxonomy (quite a hypothetical example, I admit).

I dont understand that. I read the docs again, but I cannot find what that might mean. I thought one ASV will be placed only into one tree (the one that is linked to the hmm that matched best), not several. And the taxonomic assignment of that will be reported. So how is the taxonomy of the trees relevant?
Ah now: can they have different taxonomic levels/format, i.e. when combining the taxonomic assignment based on multiple trees they might be not compatible, e.g. one is missing a taxonomic level that the other one has?

workflows/ampliseq.nf
d4straub: No I think thats fine as is.

d4straub:
Alright, looks good to me except for the unresolved comments above.

Also, I saw that
- docker | latest-everything | 1/14 = '-profile test'
- docker | latest-everything | 9/14 = '-profile test_pacbio_its'

fail with

`    > Incompatible parameters: `--sbdiexport` does not work with the chosen database of `--sintax_ref_taxonomy` because the expected taxonomic levels do not match.`
But that also appears in https://github.com/nf-core/ampliseq/pull/964 and https://github.com/nf-core/ampliseq/pull/949, i.e. its not caused by this PR here. I'll open a separate issue for that. 
edit: https://github.com/nf-core/ampliseq/issues/967
    
I am also a bit wary that the '-profile test' now takes 40min instead of ~30min. But if thats the price to pay, then lets do it.

Generally, we want to make sure that we dont break any pipeline in the next weeks regarding the strict syntax, because I expect quite soon a new stable nextflow release and this wont be compatible with the pipeline otherwise. Compatibility can be ensured with
```
NXF_VER=25.10.2 nextflow lint ampliseq -exclude ".git,.nf-test,nf-test.config"
NXF_VER=26.02.0-edge nextflow lint ampliseq -exclude ".git,.nf-test,nf-test.config"
NXF_VER=26.02.0-edge NXF_SYNTAX_PARSER=v2 nextflow run ampliseq --help
```
(where 26.02.0-edge must be the most recent version, as of today it is, afaik)

docs/usage.md
erikrikarddaniel: Forgot that when I modified this elsewhere.

modules/local/format_taxonomy.nf
erikrikarddaniel: No, it's there. The file is not among the modified anymore and in my local repo the line is back.

d4straub: Oh sorry, dont know why that happened. You are right.

nextflow_schema.json
erikrikarddaniel: > > One main reason is that not all ASVs will be placed if there are entries in the sheet without corresponding trees. (One can have entries with an hmm file that doesn't point to a tree. This is used to draw away false positive ASVs from placement in an unsuitable tree.)
> 
> But wouldnt it be theoretically fine to remove then any ASVs without taxonomic annotation from downstream analysis, because it was intentionally removed with a hmm without corresponding tree (decoy hmm)? Or, alternatively, just forward all ASVs, even without taxonomic classification, into downstream analysis, because not all downstream steps need taxonomies? Or, if both options above could be interesting, let the user decide with a param what option is happening?
> 
Personally, I'd prefer not to do this for the degree of complexity and the headache of maintenance that follows. We have a lot of params as it is and the variety of paths through the pipeline are bewildering. 

OTOH, the `if` clause that does the prioritization now is quite intimidating. An explicit param -- e.g. `--downstream_taxonomy dada2|sintax|pplace` -- would perhaps be cleaner? (If you vote for that, I'd be happy to do it, but would prefer to do that as a separate PR to make sure this works first.)

OTOH2, when I think about it, maybe we should consider splitting the pipeline into two or more: One that does denoising and various taxonomy assignments, then one or more that do the downstream? That would make it easier for the user to specify things like which taxonomy to use. Maybe this will be easier once pipeline chaining is in place. (I haven't followed what they're doing but I have a feeling this will be sophisticated.)

(One could actually imagine that, if we want to split the pipeline, output a param file that contains all the options for different params for the downstream pipeline, e.g. taxonomy, that are available given the way the upstream pipeline was run.)

In any case, I will add the current behaviour, i.e. pplace taxonomy not used, to the docs.

> > Another reason -- that we could circumvent -- is that in many cases there will be several trees and they might not have the same taxonomy (quite a hypothetical example, I admit).
> 
> I dont understand that. I read the docs again, but I cannot find what that might mean. I thought one ASV will be placed only into one tree (the one that is linked to the hmm that matched best), not several. And the taxonomic assignment of that will be reported. So how is the taxonomy of the trees relevant? Ah now: can they have different taxonomic levels/format, i.e. when combining the taxonomic assignment based on multiple trees they might be not compatible, e.g. one is missing a taxonomic level that the other one has?

Yes, what I meant was the differences in rank.

d4straub: > OTOH, the if clause that does the prioritization now is quite intimidating. An explicit param -- e.g. --downstream_taxonomy dada2|sintax|pplace -- would perhaps be cleaner? (If you vote for that, I'd be happy to do it, but would prefer to do that as a separate PR to make sure this works first.)

But that would be one more param that needs also to be checked against the taxonomies run and adds another level of complexity, in my opinion. But if you want to do that in a different PR, fine to me,, if you think that makes the code cleaner.

> maybe we should consider splitting the pipeline into two or more

Interesting idea. The pipeline is huge now, splitting might make maintenance much easier. But the pipeline is very convenient to use (at least for me), because its just one command. Once pipeline chaining is working smooth (and is not only called "usable") then I would consider that, but not before (not sure whats the state of affairs on pipeline chaining).

> Yes, what I meant was the differences in rank.

Ok that makes it complicated, I am fine as is then.

Anyway, so that comment here seems for that PR resolved then.

erikrikarddaniel:
> Alright, looks good to me except for the unresolved comments above.
> 
> Also, I saw that
> 
>     * docker | latest-everything | 1/14 = '-profile test'
> 
>     * docker | latest-everything | 9/14 = '-profile test_pacbio_its'
> 
> 
> fail with
> 
> ` > Incompatible parameters:`--sbdiexport`does not work with the chosen database of`--sintax_ref_taxonomy` because the expected taxonomic levels do not match.` But that also appears in #964 and #949, i.e. its not caused by this PR here. I'll open a separate issue for that. edit: #967
> 
OK, I don't do anything?

> I am also a bit wary that the '-profile test' now takes 40min instead of ~30min. But if thats the price to pay, then lets do it.
> 
I am too, but I didn't touch it, I believe.

We want to test everything. I can see two options here: Create a `test_quick` that only tests denosing and taxonomy assignment, and let `test` be as it is, i.e. include all the downstream. 

In line with what I suggest above -- splitting the pipeline -- one could imagine allowing input of the DADA2 tables directly into the pipeline to _only_ run the downstream parts, which would make it possible to split tests into two potentially quicker: One with the upstream, another with the downstream. It would also prepare for or substitute a split of the pipeline. IMO, there are still advantages with a split, e.g. by allowing an easier way to specify params.

> Generally, we want to make sure that we dont break any pipeline in the next weeks regarding the strict syntax, because I expect quite soon a new stable nextflow release and this wont be compatible with the pipeline otherwise. Compatibility can be ensured with
> 
> ```
> NXF_VER=25.10.2 nextflow lint ampliseq -exclude ".git,.nf-test,nf-test.config"
> NXF_VER=26.02.0-edge nextflow lint ampliseq -exclude ".git,.nf-test,nf-test.config"
> NXF_VER=26.02.0-edge NXF_SYNTAX_PARSER=v2 nextflow run ampliseq --help
> ```
> 
> (where 26.02.0-edge must be the most recent version, as of today it is, afaik)

Agreed, but I think I shouldn't do that in this PR since there are *a lot* of complaints -- warnings -- regarding code I didn't touch and this is large enough as it is. For magmap, I did this when doing the latest template update.

d4straub (approved):
Ok, I think we are through... you convinced me on all points ;)

Issue #956, opened 2026-02-25
Feature request: add ITSxRust as an alternative to ITSx for ITS extraction


ayobi:
[ITSxRust](https://github.com/ayobi/ITSxRust) is a Rust-based ITS subregion extractor that uses the same HMMER-based approach and HMM profiles as ITSx, but is designed with long-read and high-throughput workflows in mind. It's available on [Bioconda](https://bioconda.github.io/recipes/itsxrust/README.html) and could serve as an optional alternative to ITSx in ampliseq's `--cut_its` step.

### Motivation

As long-read ITS amplicon sequencing (ONT, PacBio HiFi) becomes more common, a few pain points come up with standard ITSx:

- **Throughput** — ITSxRust is written in Rust and can process large read sets faster
- **Long-read presets** — built-in `--preset ont` and `--preset hifi` flags that tune E-value thresholds and length constraints for each platform, reducing the need for manual parameter tuning
- **Partial-chain fallback** — recovers subregions when not all four ribosomal anchors are found, which is more common with noisy long reads
- **Confidence classification** — labels each extracted sequence as `confident`, `ambiguous`, or `partial`, giving downstream steps more information to filter on
- **Structured QC output** — emits a per-sample JSON summary (`--qc-json`) that could feed into MultiQC alongside existing ampliseq QC

### Proposed integration

A parameter like `--its_extractor` (defaulting to `itsx` for backward compatibility) that allows users to select `itsxrust` as an alternative. The input/output contract is the same..FASTA in, extracted ITS subregion FASTA out..so the integration should be straightforward?

ITSxRust uses the same `F.hmm` profiles from ITSx, so no additional database downloads are needed.

### References

- Repository: https://github.com/ayobi/ITSxRust
- Bioconda: `conda install -c bioconda -c conda-forge itsxrust`

Happy to help with the implementation if you guys are interested :)

d4straub:
Looks interesting!
I have myself not often ITS data, so I have no experience with those tools, but choice is always good to have, so I am supportive ;)

ayobi:
Great! I'll see about how to integrate and do a PR eventually :)

d4straub:
I found a problem with ITSxRust in CI test ["conda | 25.04.8 | 9/15"](https://github.com/nf-core/ampliseq/actions/runs/23601080996/job/68731440930?pr=976). Using conda, the pipeline fails reproducible with 
```
[...]

PR #979, opened 2026-03-30
use CONDA_PREFIX for ITSxRust F.hmm path in conda environments

ayobi:
This should fix the error with F.hmm not being found mentioned here https://github.com/nf-core/ampliseq/issues/956#issuecomment-4136319398


d4straub (approved):
I tested it, seems to work. Thanks!

ayobi:
Awesome! I'll merge :)

PR #981, opened 2026-03-30
Issue 790

### LINKED ISSUE #790: standardize the samplesheet for ampliseq to the other nf-core pipelines
### Description of feature

The samplesheet of ampliseq looks different to the other nf-core pipelines. I would propose the standardization of the headers of the samplesheet in ampliseq for:

- [ ] sampleID to sample
- [ ] forwardReads to fastq_1
- [ ] reverseReads to fastq_2
- [ ] coordinate/adapt the testdata to support the old and new samplesheet layout 

Maybe it would be good to have an adaption from tab to csv as well since with nf-core schema it already supports them.

antoniasaracco:
This closes #790


d4straub:
Hi there,
thanks for your efforts!
I had previously a look where you "only" changed the `assets/schema_input.json` to also accept the new format (`sample,fastq_1,fastq_2`). That looked good to me. Why did you now decide to go with two schemata (I dont like that solution, because it seems to complicate it a lot). What made that necessary?

antoniasaracco:
Hi @d4straub, thanks for taking a look!

The main reason is that, when I switched to the new format only, I had to modify all the input samplesheets used across the tests and update all pipeline tests accordingly.

Since everything was already working with the previous format, the idea behind accepting both formats is to avoid having to modify existing test inputs, while still supporting the new format going forward.

Let me know if you’d prefer sticking to a single format and updating everything, I can align with that approach as well.

Thanks!

d4straub:
Hi, sorry, I wasnt clear. I do like that the old and new format work.

In https://github.com/nf-core/ampliseq/pull/981/changes/4c3e99227659d746edae36dd3216f97e088c10f2 there were all (legacy & new) columns described in `assets/schema_input.json`, I liked that. However, in https://github.com/nf-core/ampliseq/pull/981/changes/ef1805ae44e4469523d49c9d825ee244063cf0e9 that change was altered so that  there is now an `assets/schema_input_new.json` and a `assets/schema_input_legacy.json` . I dont like to have two json files (duplicated maintenance), and especially not the other consequences (additions in ‎subworkflows/local/utils_nfcore_ampliseq_pipeline/main.nf‎ & workflows/ampliseq.nf) to detect and validate input schemata (whether "legacy" or "new"). So I was wondering why you changed from one `assets/schema_input.json` to `assets/schema_input_new.json` and `assets/schema_input_legacy.json`.

antoniasaracco:
hi @d4straub, thanks for clarifying!

That previous version (with a single schema_input.json using anyOf) was actually not working. I pushed it mainly to share with a collaborator and get a second look.

The issue comes from how nf-schema (specifically samplesheetToList()) validates the input. It parses the samplesheet into a list of row objects, and then validates the entire structure at once. When we tried to use anyOf at the root with two "type": "array" definitions (one per format), it caused a mismatch during validation.

In practice, this resulted in errors like:
Expected type 'object' but found type 'array'

So the “single schema with both formats” approach wasn’t viable with the current validation mechanism. Splitting into two schemas was a workaround to ensure both formats are supported without breaking validation.

I agree it’s not ideal from a maintenance perspective, but given the constraints of nf-schema, it was the most stable solution I found so far.

If you have an alternative approach in mind that keeps a single schema while working with samplesheetToList(), I’d be very happy to try it.

Thanks!

antoniasaracco:
Hi @d4straub, with the collaboration of @atrigila, we were able to implement the single schema for both formats approach. Please let us know your thoughts. Thanks!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### `Added`
 
+- ([#790](https://github.com/nf-core/ampliseq/issues/790)) - Standardized samplesheet column names to match nf-core conventions
```
d4straub:
```suggestion
- [#981](https://github.com/nf-core/ampliseq/pull/981) - Standardized samplesheet column names to match nf-core conventions (legacy samplesheet is still valid)
```

conf/test_standardized_samplesheet.config
d4straub, line 1: Is a new test really necessary? Wouldn't it be sufficient to just use the standardized samplesheet in `test.config`? Imho there are already plenty of tests, and there doesnt need to be too much overlap.

assets/schema_input.json
```diff
@@ -46,7 +65,13 @@
...
-    },
-    "uniqueEntries": ["sampleID"]
+        "anyOf": [
+            {
+                "required": ["sampleID", "forwardReads"]
+            },
+            {
+                "required": ["sample", "fastq_1"]
+            }
+        ]
```
d4straub: I am not confident here, but wouldnt that allow a samplesheet that contains the header (1) `sampleID,forwardReads,sample,fastq_1` (multiple entries for both required variables) or (2) `sampleID,forwardReads,fastq_2` (mix of legacy and standardized) or (3) `sampleID,forwardReads,reverseReads,fastq_2` (duplication of reverse read entries)?

- for (1) did you consider `oneOf` instead of `anyOf`?
- for (2) probably another nested test would be needed here? Or maybe thats fine, a mixture is here not problematic.
- for (3) maybe a `oneOf` for reverseReads,fastq_2 could work, that would require then a nesting with `allOf` for (1) and (3) I think.

assets/schema_input.json
```diff
@@ -46,7 +65,13 @@
                 "meta": ["quant_reading"]
             }
         },
-        "required": ["sampleID", "forwardReads"]
-    },
-    "uniqueEntries": ["sampleID"]
```
d4straub, line 51: There should be absolutely a check whether sampleID/sample entries are unique. If thats not possible here (I am not sure) then it could be in ampliseq.nf when parsing the samplesheet?

d4straub (commented):
I think that looks really good now! Just a few comments.

p.s.: sorry for the late reply, was in vacation.

antoniasaracco:
Hi @d4straub, Thanks a lot for your comments. 
I've implemented the new checks to not allow combinations between the old and the legacy formats in the schema.
 Also, in the workflow I added a uniqueness validation stem for the sample column. 
Then I removed the extra test added and replaced the samplesheet in test config for a one with the new format and updated accordingly the default.nf.test. Please let me know what you think.
Thanks!

docs/usage.md
```diff
@@ -118,43 +118,46 @@ Optionally, a metadata sheet can be specified for downstream analysis.
...
+The sample sheet file can be tab-separated (.tsv), comma-separated (.csv), or in YAML format (.yml/.yaml). It supports both a legacy and a standardized header layout:
 
-| Column        | Necessity | Description                                                                   |
-| ------------- | --------- | ----------------------------------------------------------------------------- |
-| sampleID      | required  | Unique sample identifiers (see below for requirements)                        |
-| forwardReads  | required  | Paths to (forward) reads zipped FastQ files                                   |
-| reverseReads  | optional  | Paths to reverse reads zipped FastQ files, required if the data is paired-end |
-| run           | optional  | If the data was produced by multiple sequencing runs, any string              |
-| control       | optional  | "control" or "sample" to allow decontamination with negative controls         |
-| quant_reading | optional  | Quantification reading to allow decontamination based on abundances           |
```
d4straub, line 130: I think that information is now completely lost? Wasn't that useful?

assets/schema_input.json
```diff
@@ -46,7 +65,41 @@
...
+                        },
+                        {
+                            "required": ["reverseReads"]
+                        }
+                    ]
+                },
+                "errorMessage": "When using standardized format (sample/fastq_1), do not use legacy fields (sampleID/forwardReads/reverseReads)"
+            }
+        ]
+    }
```
d4straub: I attempted to add here a test for the unique sample names, because it makes the pipeline fail earlier (which is desirable).
Might that work? I tested it (shortly) and it seemed to work. Could you confirm?
```suggestion
    },
    "allOf": [
        {
            "uniqueEntries": ["sample"]
        },
        {
            "uniqueEntries": ["sampleID"]
        }
    ]
```
(I first tried with "oneOf", but that didnt work as expected)
In case that seems to work for you as well, then this would replace the check for unique samples in workflows/ampliseq.nf.
In case my proposal doesnt work, then I am happy with how its now.

d4straub (commented):
Thanks, look great now!
The checks of the samplesheet worked as intended for the cases I tested.
Just two more comments, sorry for my repeated (hopefully constructive) criticism.

d4straub (approved):
Thanks, looks perfect!

Issue #915, opened 2025-10-01
Add `--sintax_ref_tax_custom`


erikrikarddaniel:
### Description of feature

To make it possible to use custom databases also for Sintax. Could be addressed together with #618 perhaps?

d4straub:
This is solved in dev!

PR #982, opened 2026-04-06
Support for custom SINTAX reference database

pieterprovoost:
Adds `--sintax_ref_tax_custom` and `--sintax_assign_taxlevels` to support use of a custom SINTAX reference database. This addresses issue #915.

## PR checklist


nextflow.config
```diff
@@ -315,6 +317,7 @@ profiles {
     test_pplace            { includeConfig 'conf/test_pplace.config'            }
     test_pplace_sheet      { includeConfig 'conf/test_pplace_sheet.config'      }
     test_sintax            { includeConfig 'conf/test_sintax.config'            }
+    test_sintax_custom     { includeConfig 'conf/test_sintax_custom.config'     }
```
d4straub: Is here an entirely new test config needed? Maybe an existing config could be changed instead? The idea would be here to minimize tests, the number of tests is already quite high imho.
Three test configs specify currently `sintax_ref_taxonomy` (test_iontorrent, test_pacbio_its, test_sintax) but only one is really reuired, I think.

nf-test.config
```diff
@@ -15,7 +15,7 @@ config {
     profile "test"
 
     // list of filenames or patterns that should be trigger a full test run
-    triggers 'nextflow.config', 'nf-test.config', 'conf/test.config', 'tests/nextflow.config', 'tests/.nftignore', 'bin/*', 'assets/schema_input.json', 'nextflow_schema.json', 'assets/report_template.Rmd'
+    triggers 'nextflow.config', 'nf-test.config', 'conf/test.config', 'conf/test_sintax_custom.config', 'tests/nextflow.config', 'tests/.nftignore', 'bin/*', 'assets/schema_input.json', 'nextflow_schema.json', 'assets/report_template.Rmd'
```
d4straub: that change isnt strictly needed (but also wouldnt hurt, but since test configs were not specified here in the past, maybe thats also not needed now?)

modules/local/format_taxonomy_sintax.nf
```diff
@@ -15,18 +15,37 @@ process FORMAT_TAXONOMY_SINTAX {
     path "versions.yml"           , emit: versions
 
     script:
-    """
-    ${params.sintax_ref_databases[params.sintax_ref_taxonomy]["fmtscript"]} \\
+    if (params.sintax_ref_tax_custom) {
+        """
+        set -- \$(head -c2 "${database}" | od -An -t u1)
+        if [ "\$#" -ge 2 ] && [ "\$1" = "31" ] && [ "\$2" = "139" ]; then
```
d4straub, line 21: Is that some sort of standard code or a custom test? If the former I could accept that, if the latter then rather not.
In any case, wouldnt it be ok to rather depend on the file ending?

d4straub (commented):
Thanks, looks quite nice! Few comments/questions.
I'll start the CI tests and see how it goes.

modules/local/format_taxonomy_sintax.nf
pieterprovoost: These are the standard first two bytes for a GZIP file as per [RFC 1952](https://www.rfc-editor.org/rfc/rfc1952). I believe a file ending check would not work for input URLs like https://ndownloader.figshare.com/files/38787078.

d4straub: Good point, then thats fine.

nf-test.config
pieterprovoost: Agreed, I have removed the test trigger.

nextflow.config
pieterprovoost: I though it would be best to test this in isolation, but I understand the concern around the number of tests. I have now merged use of a custom database into `test_iontorrent`.

d4straub: Thanks!

d4straub (approved):
Thanks, looks great!

PR #1021, opened 2026-06-23
Comparison of observed sequences to expected

d4straub:
I think it would be a good idea to have a comparison of observed versus expected results for sequences, abundances, and taxonomic assignments for mock samples and the like.

This PR adds a first step, comparing observed ASV sequences against expected reference sequence(s) and report sequence mismatches/match rates.

Included steps:
- compare generated ASV sequences to `--expected_sequences` (fasta file) with `VSEARCH --usearchglobal`
- interpret results for each sample with tables and a barplot (y = number of sequences, x = number of mismatches)
- also supplying an abundance table (tsv) with `--expected_abundances` will filter for each sample the expected sequences (without it, it is assumed that all samples have all sequences in `--expected_sequences`)

Example barplot:
b54030aeb40ed259df31ecc00c0fc8a6_2 19 0dev_sampleID_1_nucleotide-distance_barplot

Also added a CI test, data via https://github.com/nf-core/test-datasets/pull/2117

Further PRs might add in tentative priority
(1) FN/FP/TP, F1 score of (perfect) matching sequences and 
(2) evaluate how abundances compare to expected
(3) compare taxonomic classification to expected

## PR checklist


erikrikarddaniel (approved):
Looks fine, and I see the point.

It strikes me, however, that this might be better implemented as a separate pipeline that reads ampliseq output files. If a user always uses `--outdir results` the parameters could even be reused between calls, except maybe for the path to the mock/expected data. I believe this would be easier in terms of maintenance, particularly for new features of this type. (I can imagine building some sort of test on ecological distances between observed and expected mock structure.) Maybe this could one day be turned into a more general tool also applicable for e.g. metagenome and metatranscriptome mock data.

With metapipelines coming one day, calling the two pipelines would be easier too.

Happy to have in Ampliseq, and happy to discuss more.

bin/compare_sequences.r
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/env Rscript
+
+# compare_sequences.r
+
+# Get params and files from the command line
+args            <- commandArgs(trailingOnly=TRUE)
+blast6outFILE   <- args[1] # expected columns: "query+target+id+alnlen+mism+opens+qilo+qihi+tilo+tihi+ids+gaps+ql+tl+qstrand", see https://torognes.github.io/vsearch/misc/vsearch-userfields.7.html
+detabundFILE    <- args[2] # detabundFILE # calculated sequences*
```
d4straub: About naming: Here its "det" as in "detected", the comment says "calculated", further down its "res" for "result". Isnt "observed versus expected" a good naming? If yes, I could try to streamline to make that easier to follow?

bin/compare_sequences.r
```diff
@@ -0,0 +1,175 @@
...
+	print(paste("Compare file",detabundFILE,"to",expabundFILE))
+} else {
+	print(paste("Compare",detabundFILE,"to all sequences"))
+}
+
+# PREPARE
+
+# Read sequence alignment table
+blast6out = read.table( blast6outFILE, header = FALSE, sep = "\t", stringsAsFactors = FALSE, check.names = FALSE, strip.white = TRUE)
+colnames(blast6out) <- c("query","target","pident","alnlen","mismatch","gap_openings","qstart","qend","tstart","tend","ids","gaps","qlen","tlen","qstrand")
```
d4straub: the colnames do not fit the VSEARCH column names (currently "query+target+id+alnlen+mism+opens+qilo+qihi+tilo+tihi+ids+gaps+ql+tl+qstrand"), shouldnt that rather match?

bin/compare_sequences.r
```diff
@@ -0,0 +1,175 @@
...
+		print(paste("WARN - sample(s) not in pipeline output and omitted:",paste(exp_samples[!exp_samples %in% SAMPLES], collapse="," )))
+	}
+}
+
+# check if there are any samples to analyse
+if (length(SAMPLES) == 0) {
+	stop("ERROR - Found no samples to investigate")
+}
+
+# Investigate blast6out, required columns: query,target,qlen,tlen,qstart,qend,tstart,tend,gaps,mismatch
```
d4straub: The columns names here are required for the calculations. However, more are produced. Should this be minimized to the required set?

bin/compare_sequences.r
```diff
@@ -0,0 +1,175 @@
...
+blast6out$qterminalgaps <- blast6out$qlen - (abs(blast6out$qend-blast6out$qstart)+1)
+blast6out$qmismatch <- blast6out$gaps + blast6out$qterminalgaps + blast6out$mismatch
+blast6out$tterminalgaps <- blast6out$tlen - (abs(blast6out$tend-blast6out$tstart)+1)
+blast6out$tmismatch <- blast6out$gaps + blast6out$tterminalgaps + blast6out$mismatch
+if( query_or_target=="query" ) {
+	blast6out$mismatch_final <- blast6out$qmismatch
+} else if( query_or_target=="target" ) {
+	blast6out$mismatch_final <- blast6out$tmismatch
+} else if( query_or_target=="alignment" ) {
```
d4straub, line 81: I am actually not sure whether there is a difference between "query" and "alignment", since its a global alignment. The test files also do not include such a case (query is longer than target). Maybe such an expected sequence (target) should be added that is shorter than the query to make sure and to test that case?

bin/compare_sequences.r
```diff
@@ -0,0 +1,175 @@
...
+# add sample info if available
+if (file.exists(expabundFILE)) {
+	blast6out_target <- blast6out$target[blast6out$target != "*"]
+	if( !all(blast6out_target %in% exp$ID) ) {
+		print(paste("WARN - Some expected sequences in alignment results are not in",expabundFILE,". The following are missing:",paste( unique(blast6out_target[!blast6out_target %in% exp$ID]),collapse=",")))
+	}
+	if( !all(exp$ID %in% blast6out$target) ) {
+		print(paste("WARN - Some expected sequences of",expabundFILE,"are not in the alignment results. The following are missing:",paste( unique(exp$ID[!exp$ID %in% blast6out$target]),collapse=",")))
+	}
+	blast6out <- merge(blast6out,exp, by.x='target', by.y='ID', all.x=TRUE, all.y=TRUE)
```
d4straub: This is dangerous in case column names overlap. This can be if at least one sample name is identical to a column name of the VSEARCH output. The likelihood can be reduced by minimizing VSEARCH output columns and adding a test whether column names and samples overlap. Do you agree?

modules/local/compare_sequences.nf
```diff
@@ -0,0 +1,47 @@
...
+    """
+    compare_sequences.r \\
+        "$blast6out" \\
+        "$detected_abundances" \\
+        "$expected_abundances" \\
+        "" \\
+        "$similarity_threshold" \\
+        "$query_or_target" \\
+        >"nucleotide-differences.log"
+    echo "md5sum_version ${meta.id}" > "md5sum_version.txt"
```
d4straub, line 40: This file contains the md5sums of parameter minus the variable 'trace_report_suffix' (which is a time stamp). 'trace_report_suffix' is used for files in pipeline_info, such as execution report, timeline, trace, and params.json.
The idea to use the md5sum of params was to have a params-unique tag. However its also good to be able to trace that back to the params. So to complement that this file could also contain the 'trace_report_suffix'. And maybe even the params.
Or is that too much?

d4straub (commented):
Thanks @erikrikarddaniel for the review and discussion! I found a few more points, I would be happy to hear your thoughts on them, if you have the time.

> It strikes me, however, that this might be better implemented as a separate pipeline that reads ampliseq output files. If a user always uses --outdir results the parameters could even be reused between calls, except maybe for the path to the mock/expected data. I believe this would be easier in terms of maintenance, particularly for new features of this type.

Yes, that would be possibly a good idea.

bin/compare_sequences.r
erikrikarddaniel: I agree, "observed versus expected" is clearest.

bin/compare_sequences.r
d4straub: I tested it, it actually differs, so thats fine. I did not update the test files though.

modules/local/compare_sequences.nf
d4straub: I added the 'trace_report_suffix' but not the params.

bin/compare_sequences.r
erikrikarddaniel: I'm not sure I understand. In principle however I'm all for consistent column names; makes life much easier during analysis. The column names I see in the code appear to be the standard Blast names which people might be more familiar with than the Vsearch names you mention.

bin/compare_sequences.r
erikrikarddaniel: Again, only in principle sine I don't fully follow you: Any corner cases would be good to test.

bin/compare_sequences.r
erikrikarddaniel: When samples are column names this is of course a risk one should look for. Personally, I only work with "long" tables, i.e. tables with, typically, three or four columns: `sample`, `asv_id`, `count` (and `relab` if I've calculated relative abundance). This avoids the problem you see here, and makes data smaller since one can remove zeroes.

modules/local/compare_sequences.nf
erikrikarddaniel: Not exactly following again, but if a program outputs something it can be effortless. Maintenance would be the problem I'd be looking for.

bin/compare_sequences.r
d4straub: Thats a great point. I didnt consider that. I'll think about it.

d4straub: While a long format would have been a great solution, it would need some adaptions. So I omitted merging tables that could have identical column names. This way it works as well, no conflicts possible whatsoever.

d4straub:
Thanks a lot for your advice!!

Issue #1029, opened 2026-07-24
Update the pipeline to initialize channels in `PIPELINE_INITIALISATION` and more standard conversions


erikrikarddaniel:
### Description of feature

Currently, the main workflow (`workflows/ampliseq.nf`) and several local subworkflows and modules refers to parameters directly. This should be updated. Parameters that are used to create channels should be handled in `PIPELINE_INITIALISATION` in the `subworkflows/local/utils_nfcore_ampliseq_pipeline/main.nf` swf and single value parameters should be passed to the main workflow as arguments. References to parameters deeper in the code should be treated similarly.

erikrikarddaniel:
Delivered by #1032 and #1051. Remaining direct `params.*` reads elsewhere in the codebase (e.g. `modules/local/summary_report.nf`) are left as future incremental cleanup rather than tracked here.

PR #1032, opened 2026-07-24
Move self-contained channel construction into PIPELINE_INITIALISATION

erikrikarddaniel:
## PR checklist

## Description

Part of #1029: moves the params that are used only to construct a channel (with no other logic depending on how it's built) out of `workflows/ampliseq.nf` and into `PIPELINE_INITIALISATION`, threading them down through `NFCORE_AMPLISEQ` -> `AMPLISEQ` as explicit `take:`/`emit:` arguments instead of reaching into global `params` directly.

- Moves 9 channels: `metadata`, `report_template`/`report_css`/`report_logo`/`report_abstract`, `pplace_sheet`'s initial value, `expected_sequences`/`expected_abundances`, `metadata_category`.
- Pure plumbing refactor, no behavior change.
- Adds a second test case to `tests/default.nf.test` for `--report_abstract` and `--metadata_category`, the two params among these that weren't exercised by any existing test profile.
- Also includes an unrelated fix: the vendored `cutadapt` module's `stub:` block referenced `${outformat}` without ever assigning it (only `script:` did), breaking `-stub-run`.

Deliberately out of scope for this PR (left for follow-up): the remaining scalar `params.X` reads in `workflows/ampliseq.nf`, the per-classifier reference-taxonomy channels (entangled with catalog-download logic), and `params.X` reaches inside `subworkflows/local` and `modules/local`.

Verified: `nf-test` (both cases pass, no snapshot changes), `nf-core pipelines lint`, `nextflow lint .` at 26.04.0 and 26.04.6 (0 errors both), `prek run -a`.


modules/nf-core/cutadapt/main.nf
d4straub, line 1: Shouldnt that go rather into the nf-core cutadapt module? Possibly updating the module will help already (didnt check)?

tests/default.nf.test
```diff
@@ -63,4 +63,31 @@ nextflow_pipeline {
...
+        // for diversity/ancom (e.g. both "treatment1" and "mix8" beta-diversity dirs, see
+        // tests/default.nf.test.snap); restricting to "treatment1" here should exclude "mix8".
+        // (Note: alpha-rarefaction plots every metadata column regardless of --metadata_category,
+        // since QIIME2 reads the metadata file directly there, so it isn't a useful signal.)
+
+        when {
+            params {
+                outdir = "$outputDir"
+                report_abstract = "$projectDir/tower.yml"
+                metadata_category = "treatment1"
```
d4straub: I think `metadata_category` is used in `test.config` & `test_sintax.config` so not sure why that needs to be tested again.
And `report_abstract` could be used in any other test that produces a summary report, it wouldnt need an additional test, I think.

workflows/ampliseq.nf
```diff
@@ -203,29 +203,8 @@ workflow AMPLISEQ {
...
-                        refseqfile:     it.refseqfile   ? file(it.refseqfile,   checkIfExists: true) : [],
-                        refphylogeny:   it.refphylogeny ? file(it.refphylogeny, checkIfExists: true) : [],
-                        model:          it.model,
-                        taxonomy:       it.taxonomy     ? file(it.taxonomy,     checkIfExists: true) : []
-                    ]
-                ]
-            }
-    }
+    // ch_pplace_sheet: initial value built in PIPELINE_INITIALISATION from --pplace_sheet;
+    // may be overwritten below based on --dada_ref_taxonomy database config (the "pplace" key)
```
d4straub: I think I would find that comment rather in the "take:" block?

d4straub (commented):
Looks fine but I dont think the additional test is needed.
Also the cutadapt changes should be not solved here I think.

modules/nf-core/cutadapt/main.nf
erikrikarddaniel: Claude's reply: "Reviewer asked whether the outformat fix should go upstream to nf-core/modules instead of staying as a local patch. Checked upstream directly: the official module has no outformat customization at all — it's a pipeline-specific patch (confirmed the need via conf/modules.config:136's ext.outformat = "fasta"). So there's nothing to upstream; no code change, just that explanation back to the reviewer."

tests/default.nf.test
erikrikarddaniel: "The reviewer claimed it was already tested in test.config/test_sintax.config, but that was a mix-up with the different, unrelated metadata_category_barplot param. The real metadata_category (one of the 9 channels migrated into PIPELINE_INITIALISATION) had zero test coverage. Their broader point stood, though: don't add a whole new test() block just for this (doubles CI runtime). Fix: folded metadata_category = "var2" into the existing test_sintax.config profile, since it already runs ancom=true and already uses "var2" for qiime_adonis_formula, so it's a natural fit."

workflows/ampliseq.nf
erikrikarddaniel: Fixed.

erikrikarddaniel:
During work on the test updates you asked for, Claude found "a real bug in summary_report.nf — report_abstract was reading the raw params.report_abstract instead of the staged process-input variable (masked before because the old dedicated test used a local file that happened to also be a valid in-container path). That fix is included in the commit I pushed as f8b1b78d."

tests/default.nf.test
d4straub: Indeed, I was too fast and mixed it up. I like the change to use an existing test instead of adding a new one.

d4straub (approved):
LGTM!

PR #1042, opened 2026-08-04
Add per-rank DADA2 confidence columns, fix taxonomy string leak into QIIME2

### LINKED ISSUE #1041: Add more confidence values to `dada2/ASV_tax.*.tsv` and `dada2/ASV_tax_species.*.tsv`
### Description of feature

Currently, only the confidence for the most specific assigned rank is included in the output tables, but DADA2's `assignTaxonomy` method gives access to confidence for all ranks. Keep the current column but add one column for each rank, e.g. `phylum_confidence`.

erikrikarddaniel:
## PR checklist

## Description

Closes #1041.

Adds a `<rank>_confidence` column per rank to DADA2's `ASV_tax.*.tsv` and `ASV_tax_species.*.tsv`, alongside the existing overall `confidence` column. Currently only the bootstrap support for the most specific resolved rank is exported, even though `assignTaxonomy` computes it for every rank internally.

Two related fixes found while implementing this, both necessary once the new columns exist:

- `bin/parse_dada2_taxonomy.r`: the `confidence` column was already being folded into the semicolon-joined taxonomy string imported into QIIME2 (`ch_tax`) as if it were a taxonomic rank, corrupting rank-based QIIME2 operations (table filtering, ANCOM, diversity, barplot). Fixed to exclude `confidence`/`*_confidence` from that string.
- `bin/add_sh_to_taxonomy.py` (`--addsh`): derived its taxonomy rank count from a fixed column-count formula, which the new per-rank confidence columns would have silently broken. Now derives rank columns from their actual names instead.

Also closes a real CI gap found along the way: `--addsh` had **zero** test coverage anywhere in the repo (the one place it was referenced, `test_pacbio_its.config`, had it commented out, and that profile also skipped DADA2 taxonomy entirely, so it was never wired to run even uncommented). `conf/test_pacbio_its.config` now also runs DADA2 taxonomy with `--addsh` against the same UNITE database it already uses for SINTAX (`skip_dada_addspecies = true` there, since `addSpecies()` fails on these particular ITSx-extracted sequences for unrelated reasons — `addSpecies()` itself stays covered by the default 16S test).

Both `tests/default.nf.test` and `tests/pacbio_its.nf.test` pass, validated against real pipeline runs (not just lint). `prek run -a` and `nextflow lint .` clean at both the declared minimum (26.04.0) and latest Nextflow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1022](https://github.com/nf-core/ampliseq/issues/1022) - added new GloSED database for classification of Eukaryotic ITS. Not support for SH numbers as of yet. (by @tom-brekke)
 - [#1021](https://github.com/nf-core/ampliseq/pull/1021),[#1023](https://github.com/nf-core/ampliseq/pull/1023),[#1025](https://github.com/nf-core/ampliseq/pull/1025) - Comparison of observed ASVs against expected sequences or their abundance is now available with `--expected_*` parameters (by @d4straub, reviewed by @erikrikarddaniel)
 - [#1026](https://github.com/nf-core/ampliseq/pull/1026) - Add support for Oxford Nanopore Technology (ONT) R10.4 sequencing (preferably with SUP basecalling) with Savont (by @d4straub)
+- [#1041](https://github.com/nf-core/ampliseq/issues/1041) - DADA2 taxonomy tables (`ASV_tax.*.tsv`, `ASV_tax_species.*.tsv`) now include one `<rank>_confidence` column per rank, in addition to the existing overall `confidence` column (by @erikrikarddaniel)
 
 ### `Changed`
 
@@ -38,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#1019](https://github.com/nf-core/ampliseq/pull/1019) - Improve channel assignment & improve some version reporting (by @d4straub)
 - [#1027](https://github.com/nf-core/ampliseq/pull/1027) - SBDI's file `dna.tsv` recorded "paired" as lib_layout if `--single_end` was not specified alongside `--pacbio` or `--iontorrent`; now reports "single" in that cases (by @d4straub)
+- [#1041](https://github.com/nf-core/ampliseq/issues/1041) - DADA2's `confidence` column (and, before this fix, would have been every new per-rank confidence column too) was incorrectly treated as a taxonomic rank when building the taxonomy string imported into QIIME2, corrupting rank-based QIIME2 operations (table filtering, ANCOM, diversity, barplot); no longer included (by @erikrikarddaniel)
 
 ### `Dependencies`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1022](https://github.com/nf-core/ampliseq/issues/1022) - added new GloSED database for classification of Eukaryotic ITS. Not support for SH numbers as of yet. (by @tom-brekke)
 - [#1021](https://github.com/nf-core/ampliseq/pull/1021),[#1023](https://github.com/nf-core/ampliseq/pull/1023),[#1025](https://github.com/nf-core/ampliseq/pull/1025) - Comparison of observed ASVs against expected sequences or their abundance is now available with `--expected_*` parameters (by @d4straub, reviewed by @erikrikarddaniel)
 - [#1026](https://github.com/nf-core/ampliseq/pull/1026) - Add support for Oxford Nanopore Technology (ONT) R10.4 sequencing (preferably with SUP basecalling) with Savont (by @d4straub)
-- [#1041](https://github.com/nf-core/ampliseq/issues/1041) - DADA2 taxonomy tables (`ASV_tax.*.tsv`, `ASV_tax_species.*.tsv`) now include one `<rank>_confidence` column per rank, in addition to the existing overall `confidence` column (by @erikrikarddaniel)
-- [#1041](https://github.com/nf-core/ampliseq/issues/1041) - Added CI test coverage for `--addsh`, which previously had none (`test_pacbio_its` now also runs DADA2 taxonomy against the UNITE database it already uses for SINTAX) (by @erikrikarddaniel)
+- [#1042](https://github.com/nf-core/ampliseq/pull/1042) - DADA2 taxonomy tables (`ASV_tax.*.tsv`, `ASV_tax_species.*.tsv`) now include one `<rank>_confidence` column per rank, in addition to the existing overall `confidence` column (by @erikrikarddaniel)
+- [#1042](https://github.com/nf-core/ampliseq/pull/1042) - Added CI test coverage for `--addsh`, which previously had none (`test_pacbio_its` now also runs DADA2 taxonomy against the UNITE database it already uses for SINTAX) (by @erikrikarddaniel)
 
 ### `Changed`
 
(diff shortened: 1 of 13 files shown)


erikrikarddaniel:
## Why CI is failing: not a regression, a consequence of the `parse_dada2_taxonomy.r` fix

CI currently fails on 5 shards (`test`, `test_failed`, `test_multi`, `test_savont`, `test_sintax`), all with the same shape of diff: the QIIME2 barplot's snapshot expects one more level (`level-9` or `level-7`, depending on the profile) than what this branch now produces. This is expected, and here's concrete evidence of why.

`bin/parse_dada2_taxonomy.r` builds the taxonomy string QIIME2 imports (`ch_tax`) by joining every column except `ASV_ID`/`sequence`. That's the bug fixed in this PR: it was also joining in the `confidence` column, since `confidence` isn't a taxonomic rank. This script is shared across the DADA2, SINTAX, and VSEARCH-LCA branches, so the bug (and the fix) applies to all three.

Concrete before/after, using real ASV rows from this PR's own default-test run (`gtdb=R07-RS207`, `ASV_ID`s `0118d573...` and `08bdc6c8...`):

**What `dev` produces today** (the taxonomy string actually imported into QIIME2 right now):

```
0118d57337bccab8f376ba2917c68eb5    Bacteria;Bacteroidota;Ignavibacteria;Ignavibacteriales;Ignavibacteriaceae;;;;0.67
08bdc6c81f6b9daab3d132b54e1d45bc    Bacteria;;;;;;;;1
```

QIIME2's `HeaderlessTSVTaxonomyFormat` treats every `;`-separated token as a rank, positionally, regardless of content. So today, **every ASV** gets an extra trailing "rank" that's actually just its confidence score. Look at the second row: that ASV is barely classified (`Kingdom = Bacteria` only, everything else empty) — yet QIIME2 sees a non-empty value at position 9 (`"1"`), making a nearly-unclassified ASV look like it has a specific classification at some level, for any level-based operation (barplot, taxa-collapse for ANCOM, table filtering).

**With this PR's fix**, same two rows:

```
0118d57337bccab8f376ba2917c68eb5    Bacteria;Bacteroidota;Ignavibacteria;Ignavibacteriales;Ignavibacteriaceae;;;
08bdc6c81f6b9daab3d132b54e1d45bc    Bacteria;;;;;;;
```

Just the real ranks. That's why the barplot's max level correctly drops by one — it's not losing real information, it's losing a rank that was never real to begin with. I also checked what would've happened if the new per-rank `<rank>_confidence` columns (the actual point of this PR) had shipped *without* this fix: every taxonomy string would've gotten **8** bogus trailing values instead of 1, which is why fixing this was a hard requirement for this PR rather than optional cleanup.

Updating the affected snapshots (`default`, `failed`, `multi`, `savont`, `sintax`) to match the corrected output in a follow-up commit.

bin/add_sh_to_taxonomy.py
```diff
@@ -37,14 +37,22 @@
 shtax.loc[:, 9] = ""
 
 # Read taxonomy table
-# Determine number of taxonomy levels from header
-# ASV_ID  Domain  Kingdom Phylum  Class   Order   Family  Genus   confidence      sequence
+# Determine taxonomy rank columns from header -- everything except ASV_ID, sequence,
+# the aggregate "confidence" column, and the per-rank "<rank>_confidence" columns
+# (issue #1041), none of which are taxonomy ranks
```
d4straub: I wouldnt refer in the code to issues, but that might be just me.

conf/test_pacbio_its.config
```diff
@@ -34,12 +35,15 @@ params {
     cut_its = "full"
     its_extractor = "itsxrust"
 
-    // addsh = true
+    // Also runs DADA2 taxonomy (against the same UNITE database as --sintax_ref_taxonomy above)
+    // specifically to exercise --addsh, which requires a database with precomputed SH lookup
+    // files (UNITE only) and is otherwise not covered by any other test profile.
+    addsh = true
+    // dada2's addSpecies() fails on these ITSx-extracted sequences ("Non-ACGT characters"),
```
d4straub: Why are there "Non-ACGT characters"? Is that expected?

modules/local/dada2_addspecies.nf
```diff
@@ -33,9 +33,13 @@ process DADA2_ADDSPECIES {
 
     #add "Species" if not already in taxlevels
     taxlevels <- $taxlevels
+    # per-rank confidence columns (issue #1041), derived from the *original* assignTaxonomy
```
d4straub: Shall we in future add the issue number to comments in the code? Not so sure.

CHANGELOG.md
```diff
@@ -38,6 +40,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - [#1019](https://github.com/nf-core/ampliseq/pull/1019) - Improve channel assignment & improve some version reporting (by @d4straub)
 - [#1027](https://github.com/nf-core/ampliseq/pull/1027) - SBDI's file `dna.tsv` recorded "paired" as lib_layout if `--single_end` was not specified alongside `--pacbio` or `--iontorrent`; now reports "single" in that cases (by @d4straub)
+- [#1042](https://github.com/nf-core/ampliseq/pull/1042) - DADA2's `confidence` column (and, before this fix, would have been every new per-rank confidence column too) was incorrectly treated as a taxonomic rank when building the taxonomy string imported into QIIME2, corrupting rank-based QIIME2 operations (table filtering, ANCOM, diversity, barplot); no longer included (by @erikrikarddaniel)
```
d4straub: "corrupting" is by far too strong I think. It insinuates that it has a detrimental impact. To my knowledge none of these operations were impacted negatively by the confidence information.

CHANGELOG.md
```diff
@@ -38,6 +40,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- [#1042](https://github.com/nf-core/ampliseq/pull/1042) - `--addsh` derived its taxonomy rank count from a fixed column-count formula, which the new per-rank confidence columns would have silently broken; now derives it from the actual column names (by @erikrikarddaniel)
```
d4straub: That seems unnecessary? It fixes what it breaks, why is that worth mentioning?

d4straub (approved):
Generally looks good, I didnt run a test though. I assume you did :)

bin/add_sh_to_taxonomy.py
erikrikarddaniel: Removed.

modules/local/dada2_addspecies.nf
erikrikarddaniel: Better explanation added. It's actually a bug Claude found. It's making an issue for that.

CHANGELOG.md
erikrikarddaniel: Toned down.

CHANGELOG.md
erikrikarddaniel: Corrected.

conf/test_pacbio_its.config
```diff
@@ -34,12 +35,18 @@ params {
...
 
-    // addsh = true
+    // Also runs DADA2 taxonomy (against the same UNITE database as --sintax_ref_taxonomy above)
+    // specifically to exercise --addsh, which requires a database with precomputed SH lookup
+    // files (UNITE only) and is otherwise not covered by any other test profile.
+    addsh = true
+    // dada2's addSpecies() fails here: a few ASVs in this dataset become identical once
+    // trimmed to just the ITS region, so R appends ".1" to the second occurrence to keep
+    // taxtable rownames unique, and that literal ".1" then gets rejected as a non-ACGT
+    // character. Unrelated to --addsh; addSpecies itself is already covered by the default
```
d4straub, line 45: Is that an issue that sequences become identical? But thats not new introduced here, right?

erikrikarddaniel: It is, as I understand Claude, a potential R issue that likely only affects small test datasets. Not introduced here and there's an issue to fix it that I think we might as well do even if it rarely happens with real data. And, we already have a test case that exposes it once we remove this fix.

d4straub: Ok, thanks all good to me then.

PR #1045, opened 2026-08-05
Fix DADA2_ADDSPECIES crash on ITS-trimmed sequence collisions

### LINKED ISSUE #1043: DADA2_ADDSPECIES crashes with "Non-ACGT characters" when ITS-trimmed ASVs collide
### Description of the bug

`DADA2_ADDSPECIES` (dada2's `addSpecies()`) can crash with:

```
Error in assignSpecies(seqs, refFasta = refFasta, allowMultiple = allowMultiple,  :
  Non-ACGT characters present in the query sequences.
Calls: addSpecies -> assignSpecies
Execution halted
```

even though the actual ASV sequences only contain `A`/`C`/`G`/`T`.

Root cause: when `--cut_its` is used, taxonomic classification (`DADA2_TAXONOMY`/`DADA2_ADDSPECIES`) runs on the ITS-trimmed sequences rather than the full ASVs. Two or more ASVs that differ only in their flanking rDNA can become **identical once trimmed down to just the ITS region**. The taxtable built in `DADA2_TAXONOMY` (`modules/local/dada2_taxonomy.nf`) is indexed by sequence (dada2's own convention), and R requires row names to be unique — so when a duplicate ITS-trimmed sequence shows up, R silently disambiguates it by appending `.1` (then `.2`, etc.) to the row name. That row name — now containing a literal `.` and digit — is exactly what `addSpecies()` receives as a query "sequence", and its ACGT check rejects it.

Confirmed directly against a real run (`-profile test_pacbio_its,docker`, `dada_ref_taxonomy = unite-fungi=8.2`): 6 of 45 taxtable row names had a `.1` suffix; each one's un-suffixed sequence was also present as its own separate row (i.e., a genuine collision from ITS trimming, not a data artifact). Example (truncated):

```
...CCTCGGACAAACATCATTTGA          <- ASV A's row name (also the trimmed sequence)
...CCTCGGACAAACATCATTTGA.1        <- ASV B's row name, after ITS trimming collided with ASV A
```

### Impact

Any pipeline run with `--cut_its` set and `--skip_dada_addspecies` *not* set can hit this, depending on whether the specific dataset happens to produce colliding ITS-trimmed sequences. When it does, the whole `DADA2_ADDSPECIES` process fails and the run aborts, with an error message that gives no hint about the real cause (looks like a data-quality problem with the input sequences, when it isn't).

### Suggested fix

Deduplicate the ITS-trimmed sequences (or otherwise avoid relying on raw sequence identity as a row-name key) before calling `addSpecies()`, e.g. collapse taxtable rows to unique sequences before the call and map results back to all ASV_IDs sharing that sequence afterward — mirroring how `DADA2_TAXONOMY` already collapses identical full sequences via `getSequences(..., collapse = TRUE)` before `assignTaxonomy()`.

### Workaround in place today

`conf/test_pacbio_its.config` (added in #1042) sets `skip_dada_addspecies = true` specifically because this test dataset happens to trigger the crash — not a real fix, just avoids exercising the broken path in that profile's CI coverage.

### System

- nf-core/ampliseq dev branch, found while working on #1041/#1042
- Reproduced with `-profile test_pacbio_its,docker`, `dada_ref_taxonomy = unite-fungi=8.2`, `cut_its = full`, `its_extractor = itsxrust`

erikrikarddaniel:
## PR checklist

## Description

Closes #1043.

`addSpecies()` requires its taxtable's row names to be unique ACGT sequences. When `--cut_its` causes two ASVs to become identical once trimmed to just the ITS region (they differ only in flanking rDNA, which gets discarded), R silently disambiguates the duplicate row name by appending `.N` — and that literal `.N` then fails `addSpecies()`'s ACGT check, crashing the whole run with a misleading "Non-ACGT characters present in the query sequences" error.

Fix: classify each unique sequence once, then map the resulting `Species` call back onto every original row — keeping each row's own `ASV_ID` and confidence values intact (rows sharing a sequence get the same `Species` result, but nothing else is merged between them; an earlier version of this fix got that wrong and accidentally merged `ASV_ID`s too, caught by a downstream `TREESUMMARIZEDEXPERIMENT` failure on duplicate row names).

`conf/test_pacbio_its.config` previously worked around this by setting `skip_dada_addspecies = true` — that workaround is removed, so this profile is now real regression coverage for the bug. Verified the common, non-colliding case is unaffected by re-running `tests/default.nf.test` *without* `--update-snapshot` (zero diff against the existing snapshot).

### DIFF
--- a/modules/local/dada2_addspecies.nf
+++ b/modules/local/dada2_addspecies.nf
@@ -44,29 +44,14 @@ process DADA2_ADDSPECIES {
     #remove Species annotation from assignTaxonomy
     taxa_nospecies <- taxtable[,!colnames(taxtable) %in% 'Species']
 
-    # Some ASVs can end up with an identical sequence once trimmed to just the ITS
-    # region (they differ only in flanking rDNA, which gets discarded); addSpecies()
-    # requires unique ACGT row names, so classify each unique sequence once and map
-    # the result back onto every ASV that shares it.
-    is_first <- !duplicated(taxa_nospecies\$sequence)
-    taxa_unique <- taxa_nospecies[is_first, , drop = FALSE]
-    rownames(taxa_unique) <- taxa_unique\$sequence
-    taxa_unique\$sequence <- NULL
-
-    tx_unique <- addSpecies(taxa_unique, \"$database\", $args, verbose=TRUE)
-
-    # addSpecies() only adds a "Species" column; map it onto every original row by
-    # sequence, keeping each row's own original columns (ASV_ID, ranks, confidence,
-    # ...) untouched -- rows sharing a sequence get the same Species call, but
-    # nothing else is merged between them.
-    tx <- taxa_nospecies
-    tx\$Species <- tx_unique\$Species[match(taxa_nospecies\$sequence, rownames(tx_unique))]
+    tx <- addSpecies(taxa_nospecies, \"$database\", $args, verbose=TRUE)
 
     # Create a table with specified column order
+    tmp <- data.frame(row.names(tx)) # To separate ASV_ID from sequence
     expected_order <- c("ASV_ID",taxlevels,"confidence",rank_confidence_cols)
     taxa <- as.data.frame( subset(tx, select = expected_order) )
-    taxa\$sequence <- tx\$sequence
-    row.names(taxa) <- NULL
+    taxa\$sequence <- tmp[,1]
+    row.names(taxa) <- row.names(tmp)
 
     #rename Species annotation to Species_exact
     colnames(taxa)[which(names(taxa) == "Species")] <- "Species_exact"
--- a/modules/local/dada2_addspecies.nf
+++ b/modules/local/dada2_addspecies.nf
@@ -44,10 +44,30 @@ process DADA2_ADDSPECIES {
     #remove Species annotation from assignTaxonomy
     taxa_nospecies <- taxtable[,!colnames(taxtable) %in% 'Species']
 
-    tx <- addSpecies(taxa_nospecies, \"$database\", $args, verbose=TRUE)
+    # Some ASVs can end up with an identical sequence once trimmed to just the ITS
+    # region (they differ only in flanking rDNA, which gets discarded); since row
+    # names must be unique, R disambiguates the second occurrence by appending
+    # ".N" when this table is built, which addSpecies() then rejects as a
+    # non-ACGT sequence. Recover the true (possibly duplicated) sequence per row,
+    # classify each unique sequence once, then map the result back to every row
+    # that shares it.
+    true_seqs <- sub("\\\\.[0-9]+\$", "", rownames(taxa_nospecies))
+    is_first <- !duplicated(true_seqs)
+    taxa_unique <- taxa_nospecies[is_first, , drop = FALSE]
+    rownames(taxa_unique) <- true_seqs[is_first]
+
+    tx_unique <- addSpecies(taxa_unique, \"$database\", $args, verbose=TRUE)
+
+    # addSpecies() only adds a "Species" column; take that per unique sequence and
+    # map it onto every original row (including rows that shared a sequence above),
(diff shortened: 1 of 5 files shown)


modules/local/dada2_addspecies.nf
d4straub, line 1: I am not sure thats the right place for the fix. Why do that here? Why should that module fix what another module breaks? In my opinion, the optimal fix would be upstream to this module, right where the problem is caused.

erikrikarddaniel: Good point, moved it. The actual collision was caused in `dada2_taxonomy.nf`: the RDS handed off to `DADA2_ADDSPECIES` had its row names built from the query sequences themselves (assignTaxonomy's own convention), which silently breaks whenever two ASVs end up with an identical sequence (e.g. two ASVs that collapse to the same sequence once trimmed to just the ITS region) — a data.frame's row names must be unique, so R appends `.N` to the second occurrence, and that corrupted string is what `addSpecies()` later rejects as non-ACGT.

`dada2_taxonomy.nf` now exports that RDS with `ASV_ID` as row names (always unique, regardless of sequence collisions) and the actual sequence as its own explicit column. `dada2_addspecies.nf` no longer needs to regex-recover a mangled row name — it just reads the sequence column directly, classifies each unique sequence once, and maps the result back onto every ASV that shares it.

Verified this is behaviorally identical to the previous version of the fix: both `tests/pacbio_its.nf.test` (the collision case) and `tests/default.nf.test` (the common, non-colliding case) pass with zero snapshot diff against what was already committed.

erikrikarddaniel: Good that you're critical. The above was Claude wo my interference at all. I should have told ~him~ it to postpone until I had a look.

d4straub: Ok looks better.
Still some concerns: 
(1) Is the code in here (dada2_addspecies.nf) required at all? I dont think so. When removed, a few (not many I assume) sequences will be identical and this will lead to slight overhead in classification. Or would that still break? I would accept a little overhead to simplify code here.
(2) Is the problem occurring only with DADA2? Couldnt that be a wider issue after ITS sequence extraction? If yes, the fix should be rather where & when ITS is extracted, not just always, I think.

erikrikarddaniel: Both good points, and you were right to push on this — turns out the answer to (1) is yes, the extra code wasn't needed, and that also answers (2).

Traced the crash to its actual cause: #1042 added `rank_confidence` (a data.frame) into the `cbind()` in `dada2_taxonomy.nf` that builds the RDS handed to `addSpecies()`. Before #1042, that `cbind()` was vector + matrix + vector, which stays a plain matrix — and matrices tolerate duplicate row names fine. Adding a data.frame argument makes `cbind()` dispatch to `cbind.data.frame`, which forces unique row names — R silently appends `.N` to the second occurrence of a duplicate sequence, and that corrupted string is what `addSpecies()` rejects as non-ACGT.

So (1): the collapse/expand code is gone. Fix is now a one-line change — wrap `rank_confidence` in `as.matrix()` so the `cbind()` stays a matrix — and `dada2_addspecies.nf` reverts entirely to its original, pre-#1042 form. `addSpecies()` already handles genuine duplicate ACGT row names fine on its own, just classifying the duplicate sequence twice (the little overhead you mentioned). Verified both `tests/pacbio_its.nf.test` (the collision case) and `tests/default.nf.test` (the common case) pass with zero snapshot diff against the previous version of this fix — fully equivalent output, much less code.

And (2): no, it's not wider ITS-extraction issue, and this fix isn't ITS-specific either. It was never really about ITS extraction — it's a narrow consequence of one data-structure choice at one export step in `dada2_taxonomy.nf`, introduced by #1042 itself. Keeping that step a matrix protects against any source of duplicate sequences reaching `addSpecies()`, not just ITS trimming, without touching anything at the extraction step.

d4straub (approved):
Great, that looks simpler than before :D

PR #1048, opened 2026-08-07
Add comparison to taxonomic profile

d4straub:
This addition allows to supply an expected taxonomic profile that is compared to the observed taxonomic profile that is produced by the pipeline. It iterates through all common taxonomic levels and calculates a plethora of performance parameters. New parameter `--expected_profile` to supply the expected taxonomic profile.

`test_savont` was extended with a test file that was deposited at test-datasets, see https://github.com/nf-core/test-datasets/pull/2209.

Of that test, the visualisation `profile_lineplot.png` is:
profile_lineplot

## PR checklist


bin/compare_profile.r
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/env Rscript
+
+# Get params and files from the command line
+args <- commandArgs(trailingOnly=TRUE)
+if (length(args) < 3) {
+	stop("Usage: Rscript compare_profiles.r <tag> <obsFILE> <expFILE> [fbeta_val] [mode]")
```
erikrikarddaniel:
```suggestion
	stop("Usage: Rscript compare_profile.r <tag> <obsFILE> <expFILE> [fbeta_val] [mode]")
```
Script is `compare_profile.r` (singular), the usage message says `compare_profiles.r`.

bin/compare_profile.r
```diff
@@ -0,0 +1,290 @@
...
+	jaccard <- if ((tp + fp + fn) > 0) tp / (tp + fp + fn) else NA
+	return(list(recall=recall, precision=precision, f1=f1, fbeta=fbeta_res, fdr=fdr, jaccard=jaccard))
+}
+
+# Function for correlation metrics
+calculate_correlation_metrics <- function(tp_set, obs_taxa_all_lev, exp_taxa_all_lev, obs_vec, exp_vec) {
+	p_cor <- NA
+	s_rho <- NA
+	if (length(tp_set) > 1) {
+		obs_aggregated <- tapply(obs_vec, obs_taxa_all_lev, sum, NA.rm=TRUE)
```
erikrikarddaniel:
```suggestion
		obs_aggregated <- tapply(obs_vec, obs_taxa_all_lev, sum, na.rm=TRUE)
```
`NA.rm` doesn't match `sum()`'s `na.rm` parameter (R argument matching is case-sensitive), so this falls into `sum()`'s own `...` instead - `TRUE` gets coerced to `1` and silently added to the aggregated sum, and NAs are never actually dropped.

bin/compare_profile.r
```diff
@@ -0,0 +1,290 @@
...
+		exp_aggregated <- tapply(exp_vec, exp_taxa_all_lev, sum, NA.rm=TRUE)
```
erikrikarddaniel:
```suggestion
		exp_aggregated <- tapply(exp_vec, exp_taxa_all_lev, sum, na.rm=TRUE)
```
Same `NA.rm` vs `na.rm` issue as the line above.

bin/compare_profile.r
```diff
@@ -0,0 +1,290 @@
...
+	return(list(pearson_cor=p_cor, spearman_rho=s_rho))
+}
+
+# Function for distance metrics
+calculate_distance_metrics <- function(obs_taxa_at_lev, exp_taxa_at_lev, obs_taxa_all_lev, obs_vec, exp_vec) {
+	all_taxa <- union(obs_taxa_at_lev, exp_taxa_at_lev)
+	res <- list(bray_curtis=NA, hellinger=NA, jensen_shannon=NA, deviation=NA, mae=NA, rmse=NA, ps=NA)
+
+	if (length(all_taxa) > 0) {
+		obs_all_vals <- tapply(obs_vec, obs_taxa_all_lev, sum, NA.rm=TRUE)[all_taxa]
```
erikrikarddaniel:
```suggestion
		obs_all_vals <- tapply(obs_vec, obs_taxa_all_lev, sum, na.rm=TRUE)[all_taxa]
```
Same `NA.rm` vs `na.rm` issue - affects Bray-Curtis/Hellinger/Jensen-Shannon/MAE/RMSE below, all computed from this sum.

bin/compare_profile.r
```diff
@@ -0,0 +1,290 @@
...
+		exp_all_vals <- tapply(exp_vec, exp_taxa_all_lev, sum, NA.rm=TRUE)[all_taxa]
```
erikrikarddaniel:
```suggestion
		exp_all_vals <- tapply(exp_vec, exp_taxa_all_lev, sum, na.rm=TRUE)[all_taxa]
```
Same `NA.rm` vs `na.rm` issue as the line above.

erikrikarddaniel (approved):
Looks good. Claude found four real bugs: `NA.rm` where it should be `na.rm`. You will find them as suggestions in the code.

It also noticed you changed tests so that the combination of Savont and Ancomb2 is lost. Here's the message: "Checked further — ancombc2 = true is still set in conf/test.config (the default CI test, DADA2/Illumina) and conf/test_full.config, so ANCOMBC2 itself isn't losing all coverage. But test_savont.config is the only profile using sequencing_type = "nanopore" / Savont ASV-calling — so what's actually lost is coverage of ANCOMBC2 specifically against Savont-derived ASVs, a combination that had no other test exercising it.

Plausible reason (not confirmed, just inference from the diff): tax_agglom_max went from 3 → 7 in that same config, explicitly to support the new profile comparison (needs finer-grained levels exported to be meaningful). Since ANCOM-family steps typically run once per agglomeration level, keeping ancombc2 = true at tax_agglom_max = 7 instead of 3 would roughly double-plus the ANCOMBC2 runtime in that profile — dropping it may be a deliberate trade to keep CI time in check while expanding level coverage for the new feature, rather than an oversight. Worth asking the author directly rather than assuming either way."

d4straub:
Thanks for the review, all good points.
Indeed, tax_agglom_max went from 3 → 7 to get more meaningful comparisons.
ANCOMBC2 was never run anyway because no metadata was available to that test. I'll plan to improve tests in https://github.com/nf-core/ampliseq/pull/1049 though.

Issue #1033, opened 2026-07-29
Create summary tables


erikrikarddaniel:
### Description of feature

In magmap and metatdenovo we're producing summary tables in `<OUTDIR>/summary_tables` that are gzipped tsv tables with consistent column names (lower case only) ready for analysis in R, Python or whichever tool one prefers. For Ampliseq, this would mean producing two tables: `<OUTDIR>/summary_tables/ampliseq.counts.tsv.gz`, ASV counts in "long" format without zeroes, i.e. missing ASVs in a particular sample, and `<OUTDIR>/summary_tables/ampliseq.taxonomy.<tool><db>.tsv.gz`, where `<tool>` is the annotation tool, e.g. `dada_add_species` or `phyloplace`, `<db>` is the name (including version number) of the annotation database.

In an ideal world, I'd also like to consolidate taxonomies, i.e. create a table that for each ASV selects the best scoring taxonomy from a number of different databases run with the same tool. I can see two ways of doing this: Either allow the database flags, e.g. `--dada_ref_taxonomy` to take a comma-separated list of databases or let the pipeline check for existing taxonomies when invoked with `--consolidate_taxonomies`. I think the latter would be easiest if a bit unorthodox that the pipeline reads its own output directory.

I'm ready to work on a PR for this that we can take under consideration later.

erikrikarddaniel:
For quite a while, I've wanted to have the possibility to specify multiple databases as a list, e.g. `--dada_ref_taxonomy gtdb,silva,pr2`. I've discussed with my programming partner Claude ;-), and it feels like it wouldn't be that bad in terms of refactoring. A thing I'm uncertain of however, is how this will affect the QIIME2 steps. I asked Claude again: 

"""
- ch_tax_tsv (the multi-classifier channel) already coexists with multiple entries today — subworkflows/local/qiime2_export.nf:29's COMBINE_TABLE already does ch_tax_tsv.combine(...), fanning out once per classifier already present in the channel. No rework needed here; this is the channel your new taxonomy-table module would also key off, same pattern.
[...]

PR #1056, opened 2026-08-19
Support comma-separated multi-database --dada_ref_taxonomy

erikrikarddaniel:
## PR checklist

## Description

`--dada_ref_taxonomy` now accepts a comma-separated list of databases (e.g. `--dada_ref_taxonomy gtdb,silva`), running DADA2 taxonomic classification against each one. One full set of output files is published per listed database under `dada2/`.

**Scope decision, agreed with @erikrikarddaniel before implementation:** the original plan for issue #1033's remaining work bundled list-of-databases support together with consolidation (picking a single result across databases) into one PR, since list support alone leaves `ch_tax` -- the taxonomy QIIME2 filtering/diversity/barplots/ANCOM/R-objects all depend on -- ambiguous once more than one database is listed. This PR deliberately splits that: it ships list support now, with the **first-listed database** feeding every downstream step that expects exactly one taxonomy, as an explicit, documented, temporary placeholder. Consolidation (a real algorithm choice, not just "first wins") is planned as a separate, later PR. This keeps the current PR self-contained and reviewable, at the cost of a short-lived and clearly documented interim behavior.

### Implementation

Nextflow doesn't allow invoking the same process/subworkflow more than once from the same scope, so the fan-out across listed databases goes through tuple-keyed channels (`FORMAT_TAXONOMY`, `DADA2_TAXONOMY`, `DADA2_ADDSPECIES`, `FORMAT_TAXRESULTS`, `ASSIGNSH` all now take/emit a `db_key` tag) rather than a loop. `ch_tax_tsv` already tolerated multiple entries per classifier, so `COMBINE_TABLE` fans out for free once fed multiple DADA2 entries -- no changes needed there.

### Review

This branch went through two review passes before opening:
- An automated multi-file code-review pass, which caught three real edge cases (args-file collisions across databases with different `taxlevels`, a silent-misrouting bug when two listed databases resolve to identical reference files, and missing whitespace-trimming in the comma-list parsing) -- all fixed.
- A manual, file-by-file walkthrough between @erikrikarddaniel and me, which caught a design question (should databases used only for `--addsh` support be tested here? -- deferred, no meaningful 16S pair exists for it), a real gap in test coverage (winner-selection wasn't actually verified, only presence-checked -- added a content-based check using real classification data), and a genuine crash (`SUMMARY_REPORT` still indexed the raw, now-possibly-comma-joined `--dada_ref_taxonomy` param directly -- caught by an actual `nf-test` run during that review, not by inspection).

### Test plan

- [ ] `test_multidb` (new profile, two databases: `gtdb=R07-RS207,rdp=18`) passes, including content-based checks (real classification counts, not just file presence) verifying the first-listed database is actually the one used downstream
- [ ] Full regression suite passes unchanged


nextflow_schema.json
```diff
@@ -459,64 +459,14 @@
...
-                        "unite-alleuk=9.0",
-                        "unite-alleuk=8.3",
-                        "unite-alleuk=8.2",
-                        "unite-fungi",
-                        "unite-fungi=10.0",
-                        "unite-fungi=9.0",
-                        "unite-fungi=8.3",
-                        "unite-fungi=8.2",
-                        "zehr-nifh",
-                        "zehr-nifh=2.5.0"
```
d4straub, line 513: That will hurt anyone who is using a GUI (no idea how popular that is...). But there is no way around that, is there? I can only think of instead using a different parameter, but we really want to avoid that as well!

conf/test_multidb.config
```diff
@@ -0,0 +1,45 @@
...
+    // Input data
+    primer_fwd = "GTGYCAGCMGCCGCGGTAA"
+    primer_rev = "GGACTACNVGGGTWTCTAAT"
+    input = params.pipelines_testdata_base_path + "ampliseq/samplesheets/Samplesheet_v3.tsv"
+    metadata = params.pipelines_testdata_base_path + "ampliseq/samplesheets/Metadata.tsv"
+
+    // two databases, both already used elsewhere in this repo's test suite
+    dada_ref_taxonomy = "gtdb=R07-RS207,rdp=18"
+    cut_dada_ref_taxonomy = true
+
```
d4straub, line 38: This could be made a little slimmer?
skip_fastqc -> small time saving
skip_alpha_rarefaction -> large time saving
skip_diversity_indices -> medium time saving
possibly filter out a lot of ASVs to keep runtime down? E.g. by size filter or such. But maybe the number ASVs is not limiting here?

subworkflows/local/utils_nfcore_ampliseq_pipeline/main.nf
```diff
@@ -285,6 +285,14 @@ def validateInputParameters() {
         error("Incompatible parameters: Either `--skip_dada_addspecies` or `--dada_ref_tax_custom_sp` is additionally required to `--dada_ref_tax_custom`.")
     }
 
+    // --dada_ref_taxonomy defaults to a non-empty value ('sbdi-gtdb=R11-RS232-1'), so its mere presence
+    // can't distinguish "left at default" from "user explicitly asked for this (possibly a whole list)
+    // too" -- warn only when it's been changed from that default, since --dada_ref_tax_custom silently
+    // takes priority and --dada_ref_taxonomy (including every database listed in it) is ignored entirely.
+    if (params.dada_ref_tax_custom && params.dada_ref_taxonomy && params.dada_ref_taxonomy != 'sbdi-gtdb=R11-RS232-1') {
```
d4straub: Thats a good check and warning. However, I would remove the specific database, just as

```suggestion
    if (params.dada_ref_tax_custom && params.dada_ref_taxonomy) {
```
because (1) changes will be easily forgotten here in future and (2) it may very well be that the warning is helpful even when dada_ref_taxonomy is jept at default.

nextflow_schema.json
erikrikarddaniel: I agree. I would be open to have another param: `--dada_ref_databases` and keep the list for the current. The only other drawback of the list is that it has become horribly long.

conf/test_multidb.config
```diff
@@ -0,0 +1,45 @@
...
+    config_profile_description = 'Minimal test dataset to check pipeline function with a comma-separated --dada_ref_taxonomy'
+
+    // Input data
+    primer_fwd = "GTGYCAGCMGCCGCGGTAA"
+    primer_rev = "GGACTACNVGGGTWTCTAAT"
+    input = params.pipelines_testdata_base_path + "ampliseq/samplesheets/Samplesheet_v3.tsv"
+    metadata = params.pipelines_testdata_base_path + "ampliseq/samplesheets/Metadata.tsv"
+
+    // two databases, both already used elsewhere in this repo's test suite
+    dada_ref_taxonomy = "gtdb=R07-RS207,rdp=18"
```
d4straub, line 36: In my experience testing an uneven number of entries can show potential issues. I dont think its an issue here, but if that test is slimmed down a bit, maybe a third small db would be still feasible?

d4straub (approved):
Me and my AI dont have any major concern (the latter seemed sloppy though), so I think thats fine.

nextflow_schema.json
d4straub: `--dada_ref_databases` seems also fine, but would the sole upside be that we keep the enum here for `--dada_ref_database`? Maybe its worth it, maybe not, not sure.

conf/test_multidb.config
erikrikarddaniel: Test slimmed down, but kept databases at two anyway.

Issue #1061, opened 2026-08-20
Consolidate taxonomies across classification tools/methods


erikrikarddaniel:
### Description of feature

Follow-up to #1033. DADA2-only taxonomy consolidation (`--consolidate_taxonomies most-specific`/`score`, across multiple `--dada_ref_taxonomy` databases run through DADA2) is being implemented separately and deliberately scoped to stay within one classification tool. This issue is about extending consolidation *across* classification tools/methods (DADA2, SINTAX, Kraken2, QIIME2, VSEARCH-LCA, PPLACE) -- not decided yet, opened so we can work out the details in the open rather than in chat/Slack.

This was drafted with AI assistance (Claude Code), based on a design discussion working through the details below -- please push back on anything that looks wrong or oversimplified.

**Which of the two existing algorithms should even span tools?**

- `most-specific` (deepest resolved rank wins) is the more plausible candidate, since "how many ranks got resolved" is at least conceptually comparable across tools.
- `score` (highest confidence wins) should *not* span tools -- confidence isn't the same measurement across methods (DADA2's assignTaxonomy bootstrap, SINTAX's own confidence, QIIME2 classifier confidence are different metrics; VSEARCH-LCA has no confidence value at all). If `score` is active while multiple classifiers are configured, the existing classifier priority chain (SIDLE > PPLACE > DADA2 > SINTAX > Kraken2 > QIIME2 > VSEARCH-LCA) should keep deciding which *tool* wins, with `score` only refining the choice *within* that tool's own multiple databases -- exactly what the DADA2-only implementation already does today.

**Output shapes differ a lot more across tools than across DADA2's own databases**

- DADA2 / SINTAX / VSEARCH-LCA: wide, one column per rank, broadly similar -- though even within DADA2 alone we just found real rank-vocabulary differences between databases (see below), so this isn't as uniform as it first looked.
- QIIME2: single semicolon-joined `Taxon` string with rank-letter prefixes (`k__`/`p__`/`c__`/`o__`/`f__`/`g__`/`s__`), needs splitting before any rank-depth comparison is possible.
- PPLACE: single unranked semicolon-joined taxonomy string, no guaranteed depth and no rank labels at all -- genuinely unclear how to compare this against a ranked classification. Might need to just be excluded from consolidation, same as Kraken2.
[...]

Issue #1060, opened 2026-08-20
New consolidation methods: driver-table and barrnap


erikrikarddaniel:
### Description of feature

We soon have consolidation with two relatively simple methods: score and most-specific. Two other options is:

* driver-table: Let one tool/table decide which other tool/table to use. An example is when Silva can be used to decide which sequences are chloroplasts and pass them to PR2, while all other go to SBDI-GTDB. This requires a table that specifies at least which rank and taxon that directs to which tool/database.
* barrnap: Let barrnap assignment direct. This also requires a table.

Issue #1059, opened 2026-08-20
Change "Kingdom" to "Domain"


erikrikarddaniel:
### Description of the bug

Both GTDB and Silva has "Kingdom" as their top level, which is simply wrong since the content is Archaea, Bacteria or Eukaryota. Change to Domain.

(This might have historical reasons, e.g. to conform to GBIF/SBDI, but let's be correct here and let downstream tools solve this.)

@d4straub : What do you think?

### Command used and terminal output

```console

```

d4straub:
Fine with that.

PR #464, opened 2022-08-12
assignTaxonomy Species assignment

d4straub:
This PR allows Species assignment with DADA2's assignTaxonomy (results/dada2/ASV_tax.tsv) and exact match addSpecies as column "Species_exact" in (results/dada2/ASV_tax_species.tsv) for GTDB databases. This addresses https://github.com/nf-core/ampliseq/issues/351 & https://github.com/nf-core/ampliseq/issues/387.

The output changed in a way that Species assignment with assignTaxonomy will be performed and retained in case it is available from the taxonomy database. addSpecies will be run as well but not overwrite Species but instead adds a new column named Species_exact.

To run addSpecies, taxlevels (originating from --dada_assign_taxlevels or taxlevels from conf/ref_databases.config) must end with "Genus" or "Genus,Species".

Species assignment with assignTaxonomy works for databases:
- gtdb
- Silva: Using [silva_nr99_v138.1_wSpecies_train_set.fa.gz](https://zenodo.org/record/4587955/files/silva_nr99_v138.1_wSpecies_train_set.fa.gz) instead of [silva_nr99_v138.1_train_set.fa.gz](https://zenodo.org/record/4587955/files/silva_nr99_v138.1_train_set.fa.gz), only works for 138, not for 132!
- pr2

#### Can be tested for now with 
```
nextflow pull d4straub/ampliseq -r assigntaxonomy-for-species 
nextflow run d4straub/ampliseq -r assigntaxonomy-for-species -profile test,singularity --outdir result_test
```

#### Not perfectly solved:
- Unite: current file without removing species info would be fine, just the reformating script needs changes
- sbdiexport works only for databases that allow Species assignment with assignTaxonomy, because the usual databases do not report "Species" but "Species_exact" now.

## PR checklist

- [x] This comment contains a description of changes (with reason).
- [ ] If you've fixed a bug or added code that should be tested, add tests!
  - [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/ampliseq/tree/master/.github/CONTRIBUTING.md)
  - [ ] If necessary, also make a PR on the nf-core/ampliseq _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core lint`).
- [x] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Usage Documentation in `docs/usage.md` is updated.
[...]

PR #1062, opened 2026-08-20
DADA2 taxonomy consolidation across multiple --dada_ref_taxonomy databases

erikrikarddaniel:
## PR checklist

## Description

Replaces the "first-listed database wins" placeholder from #1056 with two real per-ASV consolidation
algorithms, used when `--dada_ref_taxonomy` lists more than one database:

- `--consolidate_taxonomies most-specific`: per ASV, whichever listed database resolved the deepest
  taxonomic rank wins.
- `--consolidate_taxonomies score`: per ASV, whichever listed database reported the highest
  assignTaxonomy bootstrap confidence wins.

Both break ties by declaration order in `--dada_ref_taxonomy`. Scoped to DADA2 only for now --
consolidating across different classification tools (DADA2 vs. SINTAX vs. QIIME2 etc.) is tracked
separately in #1061, and two other consolidation strategies (driver-table routing, barrnap-based) in
#1060.

### Rank vocabulary harmonization

Different databases can use different rank names for DADA2's `assignTaxonomy` taxLevels -- e.g. PR2
uses `Domain,Supergroup,Division,Subdivision,Class,...` instead of the standard
`Kingdom,Phylum,Class,...`. `most-specific` treats `Kingdom`/`Domain` and `Phylum`/`Division` as
filling the same rank slot; `Supergroup`/`Subdivision` are deliberately not counted, so a database
with more intermediate rank names doesn't win purely by having extra columns. Verified against real
downloaded reference data (see `docs/usage.md` for the full rationale) -- the note there also flags
the known asymmetry for databases with both `Kingdom` and `Domain` populated, and the Kingdom→Domain
rename itself is tracked separately in #1059.

Worth calling out: on the `test_multidb` fixture with `silva,gtdb,pr2` combined, PR2 won 71 of 347
ASVs outright (GTDB 196, SILVA 80) -- it competes meaningfully even on a bacterial-focused test
dataset, not just a token presence. Would be genuinely interesting to see how this plays out on real,
larger datasets.

### Testing

- `tests/multidb.nf.test` extended with `--consolidate_taxonomies most-specific`/`score` blocks
  (`gtdb,rdp`) and a new `silva,gtdb,pr2` block using the oldest database versions that still exercise
  PR2's `Domain/Supergroup/Division/Subdivision` scheme.
- Regression-checked against `tests/default.nf.test` and the existing single-mode `test_multidb` test.
- `nf-core pipelines lint`/`nextflow config -o json` currently fails on this branch (and on `dev`) due
  to a pre-existing, unrelated Nextflow 26.04.x bug with `SAVONT_ASV`'s closure-in-list `publishDir` --
  not something this PR introduces or can fix. `nextflow lint .` passes clean at both the declared
  minimum (26.04.0) and latest (26.04.6) Nextflow versions.


docs/usage.md
```diff
@@ -270,7 +270,9 @@ Pre-configured reference taxonomy databases are:
...
...h database resolved an ASV. A database whose own scheme has both a `Kingdom` and a separate `Domain` column populated (e.g. older PR2 releases before v5.0.0) would get credit for both, a minor and accepted asymmetry rather than something specifically guarded against. Of PR2's three extra levels between Domain and Class, only `Division` is counted, treated as filling the same slot as `Phylum`; `Supergroup` and `Subdivision` are deliberately not counted, so a database with more intermediate rank names doesn't win purely by having more columns. This mapping isn't perfectly accurate everywhere in PR2's own tree -- e.g. for the Metazoa branch, PR2's own `Class`-labelled column actually holds classical phylum names (`Arthropoda`, `Mollusca`, etc.), one level below where `Division` sits -- but `Division` is where protist phylum-equivalent groupings (`Alveolata`, `Stramenopiles`, `Rhizaria`, ...) are actually found, and protists are PR2's primary use case. A database that resolves an ASV shallower than `Division`/`Phylum` in this scheme still gets partial credit from any deeper ranks it does share names with (`Class` onward), it just isn't counted as reaching the "phylum-equivalent" level.
```
d4straub: I thinks thats a bit too much a wall-of-text regarding
- maintenance, split into several lines, will be rendered as one block, easier to see diffs
- too much details imho, too verbose, can that info be shown differently, e.g. as a table, or bullet points or such?
- some details seem not to be placed correctly here, i.e. I would not link an issue here at all (is there a precedence?) if possible to avoid it

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [#1052](https://github.com/nf-core/ampliseq/pull/1052) - Added `summary_tables/ampliseq.counts.tsv.gz`, ASV counts in long format with consistent, lower-case column names, ready for analysis in R, Python or similar without pipeline-specific parsing; also written as Parquet by default, skip with `--skip_parquet_summary` (by @erikrikarddaniel)
 - [#1056](https://github.com/nf-core/ampliseq/pull/1056) - `--dada_ref_taxonomy` now accepts a comma-separated list of databases (e.g. `gtdb,silva`), running DADA2 taxonomic classification against each; one full set of output files is published per listed database, and the first-listed database feeds every downstream step that expects a single taxonomy (consolidating multiple databases into one is planned as a future addition) (by @erikrikarddaniel)
+- [#NN](https://github.com/nf-core/ampliseq/pull/NN) - Added `--consolidate_taxonomies` (`most-specific` or `score`) to pick a per-ASV winning database across multiple `--dada_ref_taxonomy` databases, instead of always using the first-listed one; the winning result is published as `dada2/ASV_tax.consolidated.<method>.tsv` (by @erikrikarddaniel)
```
d4straub: NN = 1062 :D

d4straub (commented):
I had a first look.
Will let AI do the same later today.
Will also test `nf-core pipelines lint` again, because last time I didnt see any trouble... probably I missed something.

d4straub:
I dont really get it. I run `nf-core pipelines lint` with `nf-core/tools version 4.0.2` and `nextflow version 26.04.0.12031` on the current `dev` and there is _no_ error. How are you finding there errors?

d4straub:
So the AI review yielded:

### Issue A — `most-specific` winner-selection bias (real, partially acknowledged)
In `consolidate_dada2_taxonomy.nf` the score is `rowSums(!is.na(rank_cols))`, where `rank_cols` includes `Species` (populated by **addSpecies**, whose non-exact species calls DADA2 itself warns are unreliable) and counts **both** `Kingdom`+`Domain` and `Phylum`+`Division` (so a DB carrying both, e.g. pre-5.0.0 PR2, gets double credit). The author documented the Kingdom/Domain + Phylum/Division part as an "accepted asymmetry", but the Species part is undocumented and more consequential: a database can win purely by having emitted an often-untrustworthy species label. Separately, `score` compares raw `confidence` values that aren't normalized across DBs with different rank vocabularies/confidence scales.

-> me: This might be true, but its fine as is I think.

### Issue B — consolidated output schema drift (real compatibility risk)
The module writes a **union** of all DBs' columns — so PR2-only columns (`Domain, Supergroup, Division, Subdivision`) appear — plus a trailing `database` provenance column. Downstream consumers of `ch_dada2_tax` (`format_taxonomy.nf`, QIIME2 import, phyloseq, etc.) were written for the fixed single-DB schema `ASV_ID,<ranks>,confidence,<rank>_confidence`. The added nf-tests only assert header presence and that both DBs appear in the `database` column; they do not exercise the consolidated file through the downstream modules, so any incompatibility there is currently untested.

-> me:
The nf-test seems to cover that, the CI test `docker | 26.04.0 | 9/20` is `Test [ec87a6de] '-profile test_multidb --consolidate_taxonomies most-specific' ` and that is importing the files. However, it might be indeed that the database column leaks through but it shouldnt?
While examining the tests, I found that `docker | 26.04.0 | 8/20` is ending with 
```
       > -[nf-core/ampliseq] Pipeline completed successfully-
      PASSED (853.403s)
    Snapshots:
      Obsolete snapshots can only be checked if all tests of a file are executed successful.
```
i.e. the test passes, but the Snapshot seems to have failed? Not sure how thats possible, but that might warrant a look into it? Seems wrong.

erikrikarddaniel:
Reproduced with nf-core/tools 4.1.0 + Nextflow 26.04.6 (build 12646) -- exact error:

```
ERROR    Could not parse Nextflow config for '.../nf-core-ampliseq'. Possibly because of a Nextflow version mismatch?
         Command 'nextflow config -o json ...' failed with exit code 1
         ERROR ~ Cannot invoke method endsWith() on null object
```

`NXF_VER=25.10.0 nf-core pipelines lint` avoids it, which is what we used for this PR. You're on nf-core/tools 4.0.2 + Nextflow 26.04.0.12031 -- since that works cleanly, this might be specific to a later Nextflow 26.04.x patch (or the nf-core/tools 4.1.0 side of the `nextflow config -o json` call), not all of 26.04.x like we'd assumed. `dev` itself hits this too with our versions, so it's not something this PR introduces -- but could be worth its own upstream issue once we know exactly which version(s) are affected. Could you confirm your exact Nextflow patch?

erikrikarddaniel:
That message ("Obsolete snapshots can only be checked if all tests of a file are executed successful") is nf-test's own benign note whenever a run has any snapshot assertion mismatch within a test file -- not a sign of a real failure. Checked the actual job logs for the current (final, green) run's `8/20` and `latest-everything/8/20` shards and neither contains "obsolete" or "snapshot" at all, so whatever you saw was very likely from an earlier, now-superseded CI run during development -- already resolved by the time it went fully green (20/20 both legs).

d4straub:
> nf-core/tools 4.1.0

This pipeline is written with and based on nf-core/tools 4.0.2, it cannot be expected to succeed with 4.1.0.
Using NXF_VER=25.10.0 nf-core pipelines lint with nf-core/tools 4.1.0 is just not right.

d4straub:
> That message ("Obsolete snapshots can only be checked if all tests of a file are executed successful") is nf-test's own benign note whenever a run has any snapshot assertion mismatch within a test file -- not a sign of a real failure. Checked the actual job logs for the current (final, green) run's 8/20 and latest-everything/8/20 shards and neither contains "obsolete" or "snapshot" at all, so whatever you saw was very likely from an earlier, now-superseded CI run during development -- already resolved by the time it went fully green (20/20 both legs).

That is wrong. Check out https://github.com/nf-core/ampliseq/actions/runs/32474576241/job/96748300564?pr=1062. The issue is still there and still valid and not resolved.

erikrikarddaniel:
> > That message ("Obsolete snapshots can only be checked if all tests of a file are executed successful") is nf-test's own benign note whenever a run has any snapshot assertion mismatch within a test file -- not a sign of a real failure. Checked the actual job logs for the current (final, green) run's 8/20 and latest-everything/8/20 shards and neither contains "obsolete" or "snapshot" at all, so whatever you saw was very likely from an earlier, now-superseded CI run during development -- already resolved by the time it went fully green (20/20 both legs).
> 
> That is wrong. Check out https://github.com/nf-core/ampliseq/actions/runs/32474576241/job/96748300564?pr=1062. The issue is still there and still valid and not resolved.

It's reported as green now and I think it was this morning. In fact, everything is green so I don't understand the linting thing either.

When Claude checked this again, it found some old issue with that test which probably what you saw. But again, it's green now.

tests/multidb.nf.test
```diff
@@ -53,4 +53,81 @@ nextflow_pipeline {
...
+            // column (e.g. consolidate_dada2_taxonomy's "database", or DADA2_ADDSPECIES's
+            // "Species_exact") leaked into parse_dada2_taxonomy.r's 7-rank taxonomy string as a
+            // bogus 8th rank, same class of bug fixed in bin/parse_dada2_taxonomy.r
+            def barplot_headers = path("$outputDir/qiime2/barplot/level-7.csv").readLines()[0]
+            assertAll(
+                // real per-ASV mixing happened -- both listed databases won at least one ASV,
+                // not an accidental full-passthrough of just the first-listed one
+                { assert databases.findAll { it == "gtdb_R07-RS207" }.size() > 0 },
+                { assert databases.findAll { it == "rdp_18" }.size() > 0 },
+                { assert !path("$outputDir/qiime2/barplot/level-8.csv").exists() },
```
d4straub, line 81: this is the cause for "Obsolete snapshots..." I think. This file isnt existing, is it? See the `.snap` where it is deleted in that PR.
Edit: Oh or maybe not, because it tests that the file is _not_ existing...

erikrikarddaniel:
Two follow-ups.

## Consolidated table schema (your "does the `database` column leak through?" question)

Right instinct, worse than the `database` column. Fixed in 1f281cf8.

The module wrote the **union** of every listed database's columns -- 11 rank columns instead of 7 once PR2 is mixed with a standard-vocabulary database. But every consumer expects one vocabulary: `parse_dada2_taxonomy.r` joins all rank columns into the QIIME2 string, `phyloseq.nf` turns each into a `tax_table` rank, and `sbdiexportreannotate.R:52-54` picks `Kingdom` over `Supergroup` by presence -- with the union file both exist. A PR2-won ASV reached QIIME2 as:

```
;;Chlorophyceae;Sphaeropleales;Scenedesmaceae;;NA;Eukaryota;Archaeplastida;Chlorophyta;Chlorophyta_X
```

Empty at Kingdom and Phylum, real domain at position 8, `level-8.csv` through `level-11.csv` to match. The existing `!level-8.csv.exists()` assertion only runs in the `gtdb,rdp` test, where both databases share a vocabulary so the union adds nothing; the `silva,gtdb,pr2` test checked only the header.

The `database` column itself doesn't reach QIIME2 (excluded in 4e036f38) but does reach phyloseq as a rank -- as do `confidence`, the per-rank confidences and `sequence`, already on `dev` today.

**Fix:** harmonize each database's columns into the first-listed database's `taxlevels` before binding -- `Domain` to `Kingdom`, `Division` to `Phylum`, reverse when PR2 is first -- dropping `Supergroup`/`Subdivision`. Bootstrap columns follow their rank; `SH`/`BOLD_bin`/`Species_exact` pass through as identifiers, the distinction `sbdiexport.R:163` already makes. Scoring unchanged. Each database's own table is still published under `dada2/`.

The `silva,gtdb,pr2` test now asserts the opposite: standard ranks only, PR2-won ASVs have a populated `Kingdom`, no `level-8.csv`. All four `multidb` tests pass locally on real data.

Two pre-existing bugs turned up on the way -- an all-empty rank column becomes the literal string `NA`, and `SH` is joined in as an extra rank on `--addsh` runs. Both reproduce on clean `dev`; filed as #1075.

## "Obsolete snapshots can only be checked if all tests of a file are executed successful"

You were right and I was wrong -- real, still there, not a stale log.

It's CI sharding. `.github/actions/get-shards/action.yml` takes the shard count from `nf-test --dry-run`'s `Executed N tests`, so shards split by individual `test()` block, not by file. `tests/multidb.nf.test` has four, landing in four shards, so no shard ever runs the file whole -- and nf-test prints this whenever it runs a subset. It will appear on every run, permanently, for any file with more than one test. Running all four together locally produces no message.

So it's nf-test saying it _cannot check_ for obsolete snapshots rather than reporting a failure -- but that's a real, permanent loss of the check, not noise. Fix is to split the file, which matches your "split only when needed": this is the file that needs it. Each resulting file needs its own `removeNextflowVersion(...)` call, since `nf_test_content` lint scans whole files rather than per test block.

This PR or a separate one -- your call.

_(Investigated and implemented with help from Claude Code.)_

nextflow_schema.json
```diff
@@ -499,6 +499,12 @@
                     "help_text": "Expected amplified sequences are extracted from the DADA2 reference taxonomy using the primer sequences, that might improve classification. This is not applied to species classification (assignSpecies) but only for lower taxonomic levels (assignTaxonomy).",
                     "description": "If the expected amplified sequences are extracted from the DADA2 reference taxonomy database"
                 },
+                "consolidate_taxonomies": {
```
d4straub, line 502: so the default isnt in the enum, its just `false`? Wouldnt it be better to have "first" or such as default? That would be more explicit.

d4straub:
The summary report isnt perfect yet, the database column is leaking through (that seems the only really important point):
image
Also to see that the paragraph about extracting the database with primers (cutadapt) is only for the first database. Ideally would be both listed. Optional: a sentence how many taxonomies from what database.

Similarly for the proposed methods section (extracted by primer only one db, how many tax from what db?):
image
~~More severe might be that there seems to be only one citation?~~ Citation is fine, concatenated by "|", just didnt look close enough :(

d4straub:
I'd prefer the split of the test in that PR, because it makes it right on the first go and doesnt complicate it in my opinion.

One more consideration, but low priority: results/qiime2/rel_abundance_tables has the rel-table-ASV_with-*-tax files, if thats not too much trouble, it might be also interesting to add the consolidated taxonomy into the module that produces those files.

I run that branch with the test analysis (as above) and I really enjoyed seeing that unassigned ASVs are reduced. Also, the ASVs I looked at, had mostly overlapping taxonomies. I like that. It looks really desirable to me to use that function.

nextflow_schema.json
erikrikarddaniel: Agreed. `first` is now the explicit default and part of the enum.

erikrikarddaniel:
Thanks for testing it on real data, good to hear the unassigned ASVs go down!

Addressed:

- `--consolidate_taxonomies` now has `first` as its explicit default.
- Summary report: the `database` column no longer shows up as a taxonomic rank. A new sentence lists how many ASVs came from each database, e.g. "gtdb_R07-RS207 (315 ASVs), rdp_18 (32 ASVs)".
- Summary report and methods: when consolidating, the primer-cutting statistics are listed for every database. The methods also give the per-database ASV counts.
- `tests/multidb.nf.test` is split into one file per test.
- The consolidated taxonomy now also gets its own `qiime2/rel_abundance_tables/rel-table-ASV_with-DADA2-consolidated_<method>-tax.tsv`.

The new report assertions and the rel-table check are in `tests/multidb_mostspecific.nf.test`.

_(Implemented with help from Claude Code.)_

bin/parse_dada2_taxonomy.r
```diff
@@ -13,10 +13,11 @@ OUT="tax.tsv"
...
 
-# Join the taxonomy rank columns only -- excludes ASV_ID, sequence, the aggregate
-# "confidence" column, and the per-rank "<rank>_confidence" columns, none of which
-# are taxonomic ranks and shouldn't end up in the QIIME2 taxonomy string
-non_rank_cols <- colnames(tax) %in% c('ASV_ID', 'sequence', 'confidence') | grepl('_confidence$', colnames(tax))
+# Join the taxonomy rank columns only, excluding known non-rank columns. Matched by suffix
+# pattern where possible ("_confidence", "_exact") rather than by literal name, so a future
+# added metadata column doesn't silently leak into the taxonomy string as a bogus extra rank.
+non_rank_cols <- colnames(tax) %in% c('ASV_ID', 'sequence', 'confidence', 'database') |
+    grepl('_confidence$|_exact$', colnames(tax))
```
d4straub, line 20: Please remind me, why do we need to remove the "_exact$" here?

docs/usage.md
```diff
@@ -258,19 +258,29 @@ Pre-configured reference taxonomy databases are:
...
+- The consolidated table uses the **first-listed** database's levels throughout, whichever database won a given ASV, so that everything reading it downstream (QIIME2 import, phyloseq/TSE objects, SBDI export) sees one consistent set of ranks.
+- `Domain` is therefore written into the `Kingdom` slot and `Division` into the `Phylum` slot when the first-listed database uses the standard levels, and the reverse when it is PR2.
+- `Supergroup`/`Subdivision` have no counterpart in the standard levels and are dropped from the consolidated table. Each database's own unmodified table is still published under `dada2/`, so nothing is lost.
+- `most-specific` scores a database by how many of these levels it resolved, so a database with more intermediate rank names doesn't win purely by having more columns.
+- A database with both `Kingdom` and `Domain` populated (older PR2 releases) gets credit for both -- a minor, accepted asymmetry.
+- This mapping isn't perfect everywhere in PR2 -- e.g. Metazoa's phylum-level names sit in `Class`, one level below `Division` -- but matches PR2's primary use case (protists, whose phylum-equivalent groups are in `Division`).
```
d4straub, line 284: This is getting large and it should, important to have that explanation. And it will grow potentially.
Would it make sense to make an own section after "Taxonomic classification"? Or maybe rather a sub-section within "Taxonomic classification" (at the end)?

d4straub (approved):
I found only minor points! Great!

bin/parse_dada2_taxonomy.r
erikrikarddaniel: `Species_exact` is `DADA2_ADDSPECIES`'s own column rather than a rank: it holds the exact-match species name from `addSpecies`, kept separate from the `assignTaxonomy` approximation in `Species` since #464.

Without the exclusion it is pasted into the semicolon-joined string that QIIME2 imports, so every ASV gets an eighth level after `Species`, empty except where an exact match was found.

This is not something the PR introduces. `ASV_tax_species.<db>.tsv` has carried the column since #464, and it is `ch_dada2_tax` whenever `--skip_dada_addspecies` is not set, so the leak is live on dev and master today for any run that combines DADA2 species assignment with a QIIME2 step. I ran into it while fixing the same leak for the new `database` column and fixed both with one suffix pattern.

Happy to split it into its own PR if you would rather keep this one strictly about consolidation.

docs/usage.md
erikrikarddaniel: Agreed, it will grow. Moved to a `#### Multiple DADA2 reference databases` subsection at the end of "Taxonomic classification", with the rank vocabulary as a bullet block inside it. The database table now keeps only the sentence pointing at `conf/ref_databases.config`.

PR #1052, opened 2026-08-13
Add ampliseq.counts.tsv.gz summary table, with Parquet sibling

erikrikarddaniel:
## PR checklist

- [x] This comment contains a description of changes (with reason).
- [x] If you've fixed a bug or added code that should be tested, add tests!
- [ ] If you've added a new tool - have you followed the pipeline conventions in the [contribution docs](https://github.com/nf-core/ampliseq/tree/master/docs/CONTRIBUTING.md)
- [ ] If necessary, also make a PR on the nf-core/ampliseq _branch_ on the [nf-core/test-datasets](https://github.com/nf-core/test-datasets) repository.
- [x] Make sure your code lints (`nf-core pipelines lint`).
- [ ] Ensure the test suite passes (`nextflow run . -profile test,docker --outdir <OUTDIR>`).
- [ ] Check for unexpected warnings in debug mode (`nextflow run . -profile debug,test,docker --outdir <OUTDIR>`).
- [x] Usage Documentation in `docs/usage.md` is updated.
- [x] Output Documentation in `docs/output.md` is updated.
- [x] `CHANGELOG.md` is updated.
- [ ] `README.md` is updated (including new tool citations and authors/contributors).

## Description

First of two PRs for #1033 ("Create summary tables") — see that issue for the full multi-database/consolidation design discussion, which is entirely out of scope here.

This PR adds `summary_tables/ampliseq.counts.tsv.gz`: ASV counts in long format (`asv_id`, `sample`, `count`), zero-count rows dropped, lower-case column names throughout — matching the same convention already used in nf-core/magmap and nf-core/metatdenovo, so the output is ready to load into R, Python, etc. without pipeline-specific parsing.

Sourced directly from DADA2's own merged output (`DADA2_MERGE.out.asv`), before any of the optional post-processing filters (length, codons, decontam, etc.) run — so the table reflects raw denoising output regardless of which filters a given run enables, rather than being tied to an unclear stack of downstream choices.

A Parquet sibling (`ampliseq.counts.parquet`) is written by default via the vendored `duckdb/table2parquet` module, skippable with the new hidden `--skip_parquet_summary`.

Scoped to the DADA2 `asv_calling` path only — Savont/Nanopore ASV counts are a natural fast-follow once its `feature-table.tsv` output shape is looked into separately.

Second PR (multi-database support + taxonomy consolidation + the taxonomy summary table) will follow once the open design questions on #1033 are settled.

PR #1063, opened 2026-08-21
Add taxonomy summary table with per-ASV barrnap/decontam/filter annotations

erikrikarddaniel:
## PR checklist

## Description

Closes the last piece of issue #1033: a per-classifier taxonomy summary table, joined with per-ASV
annotations from barrnap, decontam, and the optional post-processing filters.

### `summary_tables/ampliseq.taxonomy.<classifier>.<database>.tsv.gz` (+ Parquet)

One file per classifier/database already carried in `ch_tax_tsv` (Kraken2 excluded -- its rank
vocabulary and output shape are too different to normalize). A slim, consistent-schema reformat of
that classifier's own native taxonomy table: `asv_id`, `kingdom`..`species`, `confidence`.
QIIME2's `k__`/`p__`-prefixed `Taxon` string and phylogenetic placement's unranked, semicolon-joined
string are both parsed into the same seven rank columns (best-effort for the latter). `sequence` is
dropped (recoverable via `asv_id`) and, for DADA2, the per-rank `*_confidence` columns are dropped
too (DADA2-specific detail, not comparable across classifiers) -- both remain available in full in
the native per-classifier files elsewhere in the output directory.

Joined onto every one of these files, whenever the corresponding step ran:

- `barrnap_domain`: the rRNA domain barrnap called with the lowest e-value for this ASV, blank if
  none was significant.
- `decontam_contaminant` / `decontam_not_contaminant`: decontam's own contaminant call.
- `passed_ssu_filter`, `passed_length_filter_asv`, `passed_codon_filter`,
  `passed_length_filter_itsx`: whether this ASV survived each individual optional filter.
- `ampliseq_accept`: whether this ASV survived the whole standard filtering chain (decontam through
  the ITSx-region length filter) end to end -- a convenience alongside the individual columns above,
  not instead of them, so the pipeline's own filtering choices can be recombined differently.

New module `BUILD_ASV_ANNOTATIONS` builds these by diffing each filter's ASV table immediately
before/after it runs (none of `FILTER_SSU`/`FILTER_LEN`/`FILTER_CODONS` emit a per-ASV pass/fail
column natively) -- purely additive plumbing, no change to the filter chain's own logic, order, or
gating. Tidyverse throughout (per-type long-format stacking + `pivot_wider`, split by value type
before stacking so Parquet gets real `BOOLEAN`/`UTF8` columns rather than everything coerced through
one generically-typed column).

Also fixes a gap in the existing `ampliseq.counts.tsv.gz`/`.parquet` Parquet conversion (#1052):
`DUCKDB_TABLE2PARQUET` used to run immediately after `SUMMARY_TABLE_COUNTS`, before this PR's
taxonomy tables existed yet, so it silently never converted them. Moved to run once after every
summary table has been mixed into the shared channel.

### Two real bugs found by review + testing, both fixed

- `DADA2_ADDSPECIES` always renames its own species call to `Species_exact`, only re-adding a native
  `Species` column when assignTaxonomy's own taxlevels included one -- databases that rely on
  addSpecies alone for species-level calls (e.g. RDP) would have silently had no `species` column at
  all. Fixed by coalescing `species`/`species_exact` (prefer the native assignTaxonomy call, fall
  back to addSpecies), verified against real RDP output.
- `SUMMARY_TABLE_TAXONOMY` needed `.combine()` against the annotations table -- without it, Nextflow
  silently dropped every `ch_tax_tsv` entry past the first, since a queue channel that only ever
  emits once isn't auto-broadcast against a multi-item channel the way a true value channel is.

### Testing

- `tests/default.nf.test`, `tests/multidb.nf.test`, `tests/pplace.nf.test`, `tests/sintax.nf.test`,
  `tests/vsearch_lca.nf.test` all run for real (docker), with manual content spot-checks beyond the
  automated assertions -- e.g. confirming QIIME2's `k__`/`p__` string and PPLACE's unranked string
  both parse into the right rank columns, and that the RDP `species` fix actually populates real data.
- The remaining profiles that run a non-Kraken2 classifier (`fasta`, `glosed`, `multi`, `savont`,
  `savont_independent`, `single`, `pacbio_its`, `pplace_sheet`, `qiimecustom`, `reftaxcustom`,
  `failed`, `iontorrent`) have their `.nf.test.snap` snapshots updated by hand rather than a full
  local run, based on tracing the exact classifier/database name each profile produces through the
  channel-building code (including the `user` database key used by every `*_ref_tax_custom` path).
  Flagging this explicitly in case CI surfaces a misprediction there.
- `nf-core pipelines lint` and `prek run -a` both clean, at the same baseline as `dev` (the
  pre-existing `nextflow config -o json`/SAVONT_ASV issue and a couple of unrelated `files_unchanged`
  checks aren't introduced by this PR).


workflows/ampliseq.nf
```diff
@@ -1033,6 +1050,47 @@ workflow AMPLISEQ {
         ch_qiime2_tax = channel.empty()
     }
 
+    //
+    // MODULE: Per-ASV annotations (barrnap domain call, decontam contaminant call, per-filter
+    // pass/fail, and the whole-chain "ampliseq_accept") for the taxonomy summary tables below
+    //
+    BUILD_ASV_ANNOTATIONS (
```
d4straub, line 1057: That seems like a nightmare for maintenance? Whenever a relevant tool is added it also would need to get a place here. I cannot imagine a better solution atm though. (We have one of those nightmares already: the summary report, but if the output here is similarly helpful it might be worth it.)

conf/modules.config
```diff
@@ -323,6 +323,14 @@ process {
         ]
     }
 
+    withName: SUMMARY_TABLE_TAXONOMY {
+        publishDir = [
+            path: { "${params.outdir}/summary_tables" },
+            mode: params.publish_dir_mode,
+            saveAs: { filename -> filename.equals('versions.yml') ? null : filename }
```
d4straub, line 330:
```suggestion
            saveAs: { filename -> filename?.equals('versions.yml') ? null : filename }
```
I hope that should do the linting with 4.1.0 and nextflow 26.04.0, still a bit unsure there :D

docs/output.md
```diff
@@ -211,14 +211,21 @@ DADA2 reduces sequence errors and dereplicates sequences by quality filtering, d
...
+Tables with consistent, lower-case column names, ready to load into R, Python or another tool of choice without pipeline-specific parsing.
 
 <details markdown="1">
 <summary>Output files</summary>
 
 - `summary_tables/`
-  - `ampliseq.counts.tsv.gz`: ASV counts in long format (`asv_id`, `sample`, `count`), rows with a count of 0 are dropped.
-  - `ampliseq.counts.parquet`: The same table in [Parquet](https://parquet.apache.org/) format. Skip this file with `--skip_parquet_summary`.
+  - `ampliseq.counts.tsv.gz`: ASV counts in long format (`asv_id`, `sample`, `count`), zero-count rows dropped. Sourced from DADA2's raw output, before any post-processing filters below.
+  - `ampliseq.counts.parquet`: The same table in [Parquet](https://parquet.apache.org/) format. Skip with `--skip_parquet_summary`.
```
d4straub, line 221: I would put the "Parquet" Info outside of the "details" to make it more accessible.

docs/output.md
```diff
@@ -211,14 +211,21 @@ DADA2 reduces sequence errors and dereplicates sequences by quality filtering, d
...
+  - `ampliseq.taxonomy.<classifier>.<database>.tsv.gz` (+ `.parquet`): One file pair per classifier/database actually run (Kraken2 excluded). A slim, consistent-schema reformat of that classifier's native taxonomy table: `asv_id`, `kingdom`..`species`, `confidence`. `sequence` and DADA2's per-rank `*_confidence` columns are dropped (both remain available in the native per-classifier files elsewhere in this directory).
+
+    Joined onto every file, whenever that step ran:
```
d4straub, line 224: Is this information also added to the *.counts.*  files? If not, maybe clarify.

d4straub (commented):
Didnt check the R script details, just in-person-code-browsing. Looks good so far I think. Just a few comments

workflows/ampliseq.nf
erikrikarddaniel: It is, and the shape is the same one the summary report has: a process that has to be told about every step that ran.

Concretely, a new filter costs about nine lines today, in two places.
Four in `workflows/ampliseq.nf` (a `pre`/`post` pair initialised empty, then set inside the filter's own `if` branch), and five in the module (two inputs, two `def` lines, one `add_filter_pass` call).
Nothing breaks if someone forgets: the column is simply absent, which is also what happens when the filter legitimately didn't run.
That is the part I like least, since a missed wiring looks exactly like a filter that was switched off.

The alternative I looked at is one channel of `[field_name, pre, post]` tuples instead of thirteen positional inputs, with the module looping over them.
That moves the per-filter knowledge into a single list and makes the process itself stable, but it needs each pair staged under a generated directory name and a small manifest to carry the field names, so it trades the long signature for indirection rather than removing the coupling.
I left it out because the current form is at least obvious to read.

Happy to do that refactor in this PR if you would rather have it now, or to leave it until a third or fourth filter makes the list long enough to justify the indirection.

docs/output.md
erikrikarddaniel: Moved, thanks.
The Parquet note is now in the section's prose above the collapsed block, and the individual file entries just carry a `(+ .parquet)` suffix instead of a line of their own.

docs/output.md
erikrikarddaniel: No, the annotations go onto the taxonomy tables only.
`ampliseq.counts` stays at `asv_id`, `sample`, `count`, since it is one row per ASV per sample and the annotations are per ASV.
Clarified in the docs: "Joined onto every taxonomy table, whenever that step ran, but not onto `ampliseq.counts` -- join on `asv_id` to combine them".

erikrikarddaniel: We're building a close-to normalized db structure here.

workflows/ampliseq.nf
d4straub: I see, fine as is for now I think.

workflows/ampliseq.nf
```diff
@@ -1047,6 +1064,47 @@ workflow AMPLISEQ {
...
```
d4straub: I saw in the .snap files that three tests execute BUILD_ASV_ANNOTATIONS but not SUMMARY_TABLE_TAXONOMY and produce no additional output. Additional module execution without benefit seems pointless. Maybe that could be avoided?

The three tests in question (with the settings that I suspect causing that):
- doubleprimers (skip_dada_taxonomy = true & kraken2_ref_taxonomy = "greengenes")
- multiregion (skip_dada_taxonomy = true & sidle_ref_taxonomy = "greengenes88")
- novaseq (skip_taxonomy = true)

d4straub (commented):
Looks almost good to me. Two nitpicks:
- the fruitless execution of BUILD_ASV_ANNOTATIONS (details below)
- now that we have those summery_tables folder, maybe we should also store "overall_summary.tsv" there (and rename it to maybe something like read_counts_passed.tsv)?

workflows/ampliseq.nf
erikrikarddaniel: Good, leaving it then.
It is worth revisiting once a third filter lands, since that is the point where the positional list stops being readable.

workflows/ampliseq.nf
erikrikarddaniel: Good catch, and you identified the right three.
`BUILD_ASV_ANNOTATIONS` is called unconditionally, but the only thing that reads its output is `SUMMARY_TABLE_TAXONOMY`, which consumes `ch_tax_tsv` with Kraken2 filtered out.
In all three of those profiles that filtered channel is empty -- Kraken2 only in `doubleprimers`, SIDLE never enters `ch_tax_tsv` in `multiregion`, and `novaseq` skips taxonomy altogether -- so the annotations get built and then dropped.

I would rather not gate it on the parameters, since `ch_tax_tsv` is filled from six separate places under six different conditions and a parameter expression would have to restate all of them, which is the maintenance trap you pointed at in the other thread.
Gating on the channel itself keeps the two in step:

```groovy
ch_summary_tax = ch_tax_tsv.filter { meta, _tsv -> meta.classifier != "KRAKEN2" }

ch_annot_barrnap = ch_summary_tax
    .map { _meta, _tsv -> true }
    .first()
    .combine( ch_barrnapsummary.ifEmpty( [] ) )
    .map { _gate, summary -> summary }
```

The gated channel replaces the module's first input and the other twelve stay as they are, so the module itself does not change.
I checked the operator behaviour on its own before wiring it in: an empty taxonomy channel leaves the process unexecuted, and several taxonomy tables still run it exactly once.

I will push this and confirm the three snapshots lose the process without anything else moving.

erikrikarddaniel: Done in ecd01159.

`BUILD_ASV_ANNOTATIONS` now takes its first input through a gate derived from the channel `SUMMARY_TABLE_TAXONOMY` consumes, so it runs only when there is a taxonomy table to join the annotations onto.
The module itself is unchanged, and `SUMMARY_TABLE_TAXONOMY` reuses the same filtered channel rather than filtering a second time.

All three profiles you named now lose the module's entry from the versions file and nothing else moves in their snapshots:

- `doubleprimers` passed
- `multiregion` passed
- `novaseq` passed

`fasta` was run as a control, since it does write a summary table, and its snapshot is byte-identical.

d4straub (approved):
LGTM!

PR #1065, opened 2026-08-26
Add --max_ee_r to set DADA2 maxEE separately for reverse reads

### LINKED ISSUE #1037: Separate maxEE for paired-end forward and reverse reads
### Description of feature

Hi!

When running `DADA2` as a stand-alone pipeline in` R` on paired-end reads, I am able to specify a vector with two different values for `maxEE` during the `filterAndTrim` step - one for forward reads, and one for reverse reads. With `nf-core/ampliseq`, I can only set a single integer for the `--max_ee` parameter, applied to both forward and reverse reads simultaneously.

I have a data set with reads from several sequencing runs. For some runs, the Illumina NextSeq data has some quality drops along the entire read length of the reverse reads only (apparently a strange but known behaviour with this platform). I am specifying different `--trunclenf` and `--trunclenr` values, but I would also like to relax `--max_ee` for reverse reads only. 

Seeing that vectors of integers can be set for example for the `--binned_quality` paramater, I don't see why this wouldn't be possible for `--max_ee` as well?

Just a thought! 

Best
Nauras

erikrikarddaniel:
## PR checklist

## Description

Closes #1037.

On paired-end Illumina data, `--max_ee` currently applies to both forward and reverse reads via one shared `filterAndTrim(maxEE = c(x, x))` call. Some sequencing runs show a quality drop restricted to the reverse read only (a known Illumina NextSeq behaviour, per the issue), for which a single shared threshold forces an unwanted trade-off between the two reads.

This adds `--max_ee_r`, which lets the reverse read use its own, typically more permissive, expected-error threshold instead — mirroring the existing `--trunclenf`/`--trunclenr` precedent for separate forward/reverse truncation lengths.

- `--max_ee_r` unset (default): behaviour is unchanged. `conf/modules.config`'s new ternary renders byte-identical `ext.args` output to before, and `default.nf.test`'s snapshot needed no update.
- `--max_ee_r` set: overrides `--max_ee` for the reverse read only, for paired-end Illumina data (the only sequencing type with a separate reverse read in the first place).
- `modules/local/summary_report.nf` / `assets/report_template.Rmd`: the report now states forward/reverse thresholds separately when `--max_ee_r` is set, instead of always reporting one shared value.

### Test coverage

Added a small, dedicated `test_max_ee_r` profile (`conf/test_max_ee_r.config` + `tests/max_ee_r.nf.test`) rather than a module-level `nf-test` for `DADA2_FILTNTRIM`. This repo's local modules are flat single-file, and a module-level test would require converting to the nested `main.nf`/`meta.yml`/`tests` layout that only vendored `modules/nf-core/**` components use — and this repo's own `nf-test.config` explicitly ignores tests under that path, so it wouldn't even run in CI. The new profile skips taxonomy and QIIME2 entirely (not needed to exercise DADA2 read filtering), keeping it one of the cheaper test profiles in the suite (~2.5 min locally), and asserts directly on the published `dada2/args/filterAndTrim.args.txt` content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - [#1042](https://github.com/nf-core/ampliseq/pull/1042) - Added CI test coverage for `--addsh`, which previously had none (`test_pacbio_its` now also runs DADA2 taxonomy against the UNITE database it already uses for SINTAX) (by @erikrikarddaniel)
 - [#1052](https://github.com/nf-core/ampliseq/pull/1052) - Added `summary_tables/ampliseq.counts.tsv.gz`, ASV counts in long format with consistent, lower-case column names, ready for analysis in R, Python or similar without pipeline-specific parsing; also written as Parquet by default, skip with `--skip_parquet_summary` (by @erikrikarddaniel)
 - [#1056](https://github.com/nf-core/ampliseq/pull/1056) - `--dada_ref_taxonomy` now accepts a comma-separated list of databases (e.g. `gtdb,silva`), running DADA2 taxonomic classification against each; one full set of output files is published per listed database, and the first-listed database feeds every downstream step that expects a single taxonomy (consolidating multiple databases into one is planned as a future addition) (by @erikrikarddaniel)
+- [#NN](https://github.com/nf-core/ampliseq/pull/NN) - New `--max_ee_r` overrides `--max_ee` for reverse reads only (paired-end Illumina data), for runs where read quality drops on the reverse read only (fixes [#1037](https://github.com/nf-core/ampliseq/issues/1037)) (by @erikrikarddaniel)
 
 ### `Changed`
 
--- a/assets/report_template.Rmd
+++ b/assets/report_template.Rmd
@@ -34,6 +34,7 @@ params:
     trunclenf: ""
     trunclenr: ""
     max_ee: ""
+    max_ee_r: ""
     trunc_qmin: FALSE
     trunc_rmin: ""
     dada_sample_inference: ""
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - [#1052](https://github.com/nf-core/ampliseq/pull/1052) - Added `summary_tables/ampliseq.counts.tsv.gz`, ASV counts in long format with consistent, lower-case column names, ready for analysis in R, Python or similar without pipeline-specific parsing; also written as Parquet by default, skip with `--skip_parquet_summary` (by @erikrikarddaniel)
 - [#1056](https://github.com/nf-core/ampliseq/pull/1056) - `--dada_ref_taxonomy` now accepts a comma-separated list of databases (e.g. `gtdb,silva`), running DADA2 taxonomic classification against each; one full set of output files is published per listed database, and the first-listed database feeds every downstream step that expects a single taxonomy (consolidating multiple databases into one is planned as a future addition) (by @erikrikarddaniel)
+- [#NN](https://github.com/nf-core/ampliseq/pull/NN) - New `--max_ee_r` overrides `--max_ee` for reverse reads only (paired-end Illumina data), for runs where read quality drops on the reverse read only (fixes [#1037](https://github.com/nf-core/ampliseq/issues/1037)) (by @erikrikarddaniel)
```
d4straub: Needs actual PR number

d4straub (approved):
Looks fine, except the CHANGELOG that needs an update.

I am not a big fan of the additional test. I do like that to be tested, but not with an extra not-so-cheap test (~4 min on CI). I suggest either to add `mex_ee_r` instead to one of the existing Illumina PE test configs or generate something like the tiny t`ests/taxonomy_hash.nf.test`. While I would prefer that solutions, I dont want to block the PR because of that.

erikrikarddaniel:
> Looks fine, except the CHANGELOG that needs an update.
> 
I always forget. Claude has become better at remembering -- much better than me -- recently though. This was probably before that. Done.

> I am not a big fan of the additional test. I do like that to be tested, but not with an extra not-so-cheap test (~4 min on CI). I suggest either to add `mex_ee_r` instead to one of the existing Illumina PE test configs or generate something like the tiny t`ests/taxonomy_hash.nf.test`. While I would prefer that solutions, I dont want to block the PR because of that.

Good point. Being done.

d4straub:
I dont understand what you did there. Thats still an additional test, just not an additional file. I think that last commit made it actually worse than better.  Thats worse in my opinion because it hides the test in a config for another test but still requires a similar runtime. It was not about the file count (i.e. merging .test files), it was about the number of tests. I suggested to change any .config with Illumima PE data so that it contains `max_ee_r = 5` or such.
Sorry for having not made myself clear earlier.

erikrikarddaniel:
> I suggested to change any .config with Illumima PE data

Yes, I should have checked more carefully. We're adding it to `test_novaseq` now. Will be another PR I assume since I merged this yesterday.

